PR #67, opened 2023-06-19
fix: internal audit comments

milapsheth:


contracts/interchainTokenService/InterchainTokenService.sol
```diff
@@ -322,6 +323,8 @@ contract InterchainTokenService is IInterchainTokenService, AxelarExecutable, Up
         _deployRemoteStandardizedToken(tokenId, name, symbol, decimals, distributor, admin, destinationChain, gasValue);
     }
 
+    // TODO: pause check isn't needed for this since it's just passing tokens through, so we can save some gas
+    // TODO: Add commandID to help express vs execute reorder protection, i.e check if commandID is already executed
```
Foivos, line 327: Are we ok with the wait to generate a commandID for express? If we know a commandID can't we just execute?

contracts/interchainTokenService/InterchainTokenService.sol
```diff
@@ -49,15 +49,16 @@ contract InterchainTokenService is IInterchainTokenService, AxelarExecutable, Up
...
+    bytes32 internal constant PREFIX_STANDARDIZED_TOKEN_ID = keccak256('its-standardized-token-id');
+    bytes32 internal constant PREFIX_STANDARDIZED_TOKEN_SALT = keccak256('its-standardized-token-salt');
 
     uint256 private constant SELECTOR_SEND_TOKEN = 1;
     uint256 private constant SELECTOR_SEND_TOKEN_WITH_DATA = 2;
     uint256 private constant SELECTOR_DEPLOY_TOKEN_MANAGER = 3;
     uint256 private constant SELECTOR_DEPLOY_AND_REGISTER_STANDARDIZED_TOKEN = 4;
 
-    // keccak256('interchain-token-service')-1
+    // TODO: For contract Id, we can stick to not using the -1 convention
```
Foivos, line 60: But why?

contracts/interchainTokenService/InterchainTokenService.sol
```diff
@@ -337,6 +340,11 @@ contract InterchainTokenService is IInterchainTokenService, AxelarExecutable, Up
         _setExpressSendToken(tokenId, destinationAddress, amount, sendHash, caller);
     }
 
+    // TODO: express receive with data should be opt-in since the data is opaque and might be high value
```
Foivos, line 343: Will do

contracts/interchainTokenService/InterchainTokenService.sol
```diff
@@ -381,6 +389,8 @@ contract InterchainTokenService is IInterchainTokenService, AxelarExecutable, Up
         bytes calldata destinationAddress,
         uint256 amount
     ) external payable onlyTokenManager(tokenId) notPaused {
+        // TODO: what's the point of sendHash? block.number can't always enforce uniqueness
```
Foivos, line 393: It is to make tracking a bit easier and add the possibility to express two different but similar calls that arrive close to each other (as long as they are at least a block appart)

contracts/interfaces/IInterchainTokenExecutable.sol
```diff
@@ -1,14 +1,19 @@
...
 
+// TODO: Add a base implementation that requires caller of execute to be the token service
+// All interfaces should be documented, especially since external devs look at interfaces first
 // General interface for upgradable contracts
 interface IInterchainTokenExecutable {
-    function exectuteWithInterchainToken(
+    // TODO: typo
+    function executeWithInterchainToken(
         string calldata sourceChain,
+        // TODO: If we really want to keep the interface similar, we can use string here but cast it to bytes internally
```
Foivos, line 13: I think that would be too confusing to developers. Keeping an interface the same but changing the functionality is much worse than changing both.

contracts/utils/Multicall.sol
```diff
@@ -5,14 +5,17 @@ pragma solidity 0.8.9;
 import { IMulticall } from '../interfaces/IMulticall.sol';
 
 contract Multicall is IMulticall {
+    error MulticallFailed(bytes err);
```
Foivos, line 8: I don't like doing this because then the error will not easily be decodable by from the interface.

contracts/interchainTokenService/InterchainTokenService.sol
milapsheth: commandID is currently known when the contract call is made. It might still get reorged, but it's essentially hash of tx hash + event index + chain id. So it still helps protect in normal scenarios

contracts/interchainTokenService/InterchainTokenService.sol
milapsheth: forgot to add that we are doing it for our other contracts, so for consistency. Also, in this case the hash isn't used for a storage slot, so -1 benefit is a bit orthogonal

contracts/interchainTokenService/InterchainTokenService.sol
milapsheth: i'm thinking that we can use commandId for it instead

contracts/utils/Multicall.sol
milapsheth: i see, fair enough then

contracts/interfaces/IInterchainTokenExecutable.sol
milapsheth: yeah, not seeing much benefit too

PR #69, opened 2023-06-21
feat: negative test coverage

deanamiel:
PR for negative test coverage for the interchain token service


test/tokenService.js
```diff
@@ -295,6 +473,7 @@ describe('Interchain Token Service', () => {
             expect(await tokenManager.admin()).to.equal(admin);
         });
 
+        // Isn't this the same test as above?
         it('Should be able to receive a remote standardized token depoloyment with a mint/burn token manager', async () => {
```
milapsheth, line 477: looks like it, can you remove it?

test/tokenService.js
```diff
@@ -422,11 +677,29 @@ describe('Interchain Token Service', () => {
                 .and.to.emit(gateway, 'ContractCall')
                 .withArgs(service.address, chain, service.address.toLowerCase(), keccak256(payload), payload);
         });
+
+        it('Should revert on remote custom token manager deployment if paused', async () => {
```
milapsheth, line 681: this should be a separate describe test so you don't have to set it to unpause for every other test

Foivos: will maybe fix this later, noted

PR #74, opened 2023-06-23
feat: addresses almost all comments from the 23.6.2023 report

Foivos:


contracts/interchain-token/InterchainToken.sol
```diff
@@ -62,7 +61,10 @@ abstract contract InterchainToken is IInterchainToken, ERC20 {
          * @dev if you know the value of `tokenManagerRequiresApproval()` you can just skip the if statement and just do nothing or _approve.
          */
         if (tokenManagerRequiresApproval()) {
-            _approve(sender, address(tokenManager), allowance[sender][address(tokenManager)] + amount);
+            uint256 allowance_ = allowance[sender][address(tokenManager)];
+            if(allowance_ != type(uint256).max) {
+                _approve(sender, address(tokenManager), allowance_ + amount);
+            }
```
milapsheth, line 67:
```suggestion
            if (allowance_ > type(uint256).max - amount) {
                allowance_ = type(uint256).max - amount
            }
            
            _approve(sender, address(tokenManager), allowance_ + amount);
```
maybe something like this so it's more robust against overflows?

contracts/libraries/AddressBytesUtils.sol
```diff
@@ -4,16 +4,18 @@ pragma solidity 0.8.9;
...
 library AddressBytesUtils {
+    error InvalidBytesLength(bytes bytesAddress);
+
     /**
      * @dev Converts a bytes address to an address type.
      * @param bytesAddress The bytes representation of an address
      * @return addr The converted address
      */
     function toAddress(bytes memory bytesAddress) internal pure returns (address addr) {
+        if(bytesAddress.length != 20) revert InvalidBytesLength(bytesAddress);
```
milapsheth, line 18:
```suggestion
        if (bytesAddress.length != 20) revert InvalidBytesLength(bytesAddress);
```

contracts/utils/ExpressCallHandler.sol
```diff
@@ -16,8 +15,11 @@ contract ExpressCallHandler is IExpressCallHandler {
     uint256 internal constant PREFIX_EXPRESS_RECEIVE_TOKEN = 0x67c7b41c1cb0375e36084c4ec399d005168e83425fa471b9224f6115af865619;
     // uint256(keccak256('prefix-express-give-token-with-data'));
     uint256 internal constant PREFIX_EXPRESS_RECEIVE_TOKEN_WITH_DATA = 0x3e607cc12a253b1d9f677a03d298ad869a90a8ba4bd0fb5739e7d79db7cdeaad;
-    mapping(bytes32 => address) private expressGiveToken;
-    mapping(bytes32 => address) private expressGiveTokenWithData;
+
+    modifier onlyDifferentCaller(address destinationAddress, address expressCaller) {
```
milapsheth, line 19: why restrict this? express caller might be using the service to send tokens itself, while operating the express relayer. ERC20 transfer should still go through fine

milapsheth: Saw the report comment, we should use the original `amount` arg for setting the express caller in storage. We don't need to do the balance difference for the purpose of the event (we did genuinely receive x tokens, after they're transferred to the recipient, any fees charged is not our concern).

And in fact, when the normal call is received, we're using the original amount. So, atm we won't be reading the express storage slot correctly if it was a fee-on-transfer token.

milapsheth (approved):

contracts/interchain-token/InterchainToken.sol
Foivos: That's not even checked in the ERC20 implementation as type(uint256).max is considered too large.

contracts/interchain-token-service/InterchainTokenService.sol
```diff
@@ -489,6 +483,12 @@ contract InterchainTokenService is IInterchainTokenService, AxelarExecutable, Up
     INTERNAL FUNCTIONS
     \****************/
 
+    function _sanitizeTokenManagerImplementation(address[] memory implementaions, uint256 i) internal pure returns (address implementation) {
```
milapsheth, line 486:
```suggestion
    function _sanitizeTokenManagerImplementation(address[] memory implementations, uint256 i) internal pure returns (address implementation) {
```

contracts/interchain-token/InterchainToken.sol
milapsheth: Hmm, but in our case, we're adding to an existing value vs just overwriting to a provided uint256. Not a strong opinion though

contracts/interchain-token-service/InterchainTokenService.sol
milapsheth: Also pass the token manager type instead of an int?

PR #87, opened 2023-07-11
feat: least autority fixes

Foivos:


contracts/utils/Operatable.sol
```diff
@@ -51,4 +53,28 @@ contract Operatable is IOperatable {
...
+
+    /**
+     * @notice Proposed a change of the operator of the contract
+     * @dev Can only be called by the current operator
+     * @param operator_ The address of the new operator
+     */
+    function proposeOperatorChange(address operator_) external onlyOperator {
+        assembly {
+            sstore(PROPOSED_OPERATOR_SLOT, operator_)
+        }
```
milapsheth, line 65: add an event for operator transfer started. same for distributor

contracts/utils/Distributable.sol
```diff
@@ -51,4 +54,28 @@ contract Distributable is IDistributable {
...
+
+    /**
+     * @notice Proposed a change of the distributor of the contract
+     * @dev Can only be called by the current distributor
+     * @param distr The address of the new distributor
+     */
+    function proposeDistributorChange(address distr) external onlyDistributor {
+        assembly {
+            sstore(PROPOSED_DISTRIBUTOR_SLOT, distr)
+        }
```
milapsheth, line 66:
```suggestion
    function proposeDistributorChange(address distributor) external onlyDistributor {
        assembly {
            sstore(PROPOSED_DISTRIBUTOR_SLOT, distributor)
        }
```

contracts/interfaces/IDistributable.sol
```diff
@@ -19,4 +21,17 @@ interface IDistributable {
      * @param distributor The address of the new distributor
      */
     function setDistributor(address distributor) external;
```
milapsheth, line 23: Maybe switch to `transfer` instead of `set` for consistency

Foivos: `transferOwnership` => `transferDistributorship` sounds weird we could maybe do `transferDistributor` but that is grammatically incorrect so idk

PR #92, opened 2023-07-24
feat: gas optimizations

Foivos:


contracts/remote-address-validator/RemoteAddressValidator.sol
```diff
@@ -27,7 +31,17 @@ contract RemoteAddressValidator is IRemoteAddressValidator, Upgradable {
     constructor(address _interchainTokenServiceAddress) {
         if (_interchainTokenServiceAddress == address(0)) revert ZeroAddress();
         interchainTokenServiceAddress = _interchainTokenServiceAddress;
-        interchainTokenServiceAddressHash = keccak256(bytes(_lowerCase(interchainTokenServiceAddress.toString())));
+        string memory interchainTokenServiceAddressString = interchainTokenServiceAddress.toString();
+        interchainTokenServiceAddressHash = keccak256(bytes(_lowerCase(interchainTokenServiceAddressString)));
```
milapsheth, line 35:
```suggestion
        interchainTokenServiceAddressHash = keccak256(bytes(nterchainTokenServiceAddressString));
```
since the string comes from `toString`, don't need to lowercase it again?

contracts/remote-address-validator/RemoteAddressValidator.sol
```diff
@@ -16,8 +16,12 @@ contract RemoteAddressValidator is IRemoteAddressValidator, Upgradable {
     mapping(string => string) public remoteAddresses;
     address public immutable interchainTokenServiceAddress;
     bytes32 public immutable interchainTokenServiceAddressHash;
+    uint256 private immutable interchainTokenServiceAddress1;
```
milapsheth, line 19: add a @dev comment explaining how this works

contracts/remote-address-validator/RemoteAddressValidator.sol
```diff
@@ -16,8 +16,12 @@ contract RemoteAddressValidator is IRemoteAddressValidator, Upgradable {
...
+    uint256 private immutable interchainTokenServiceAddress2;
     mapping(string => bool) public supportedByGateway;
 
+    
+
```
milapsheth, line 24:
```suggestion
```

contracts/remote-address-validator/RemoteAddressValidator.sol
```diff
@@ -27,7 +31,17 @@ contract RemoteAddressValidator is IRemoteAddressValidator, Upgradable {
...
+        interchainTokenServiceAddressHash = keccak256(bytes(_lowerCase(interchainTokenServiceAddressString)));
+        uint256 p1;
+        uint256 p2;
+        assembly {
+            p1 := mload(add(interchainTokenServiceAddressString, 32))
+            p2 := mload(add(interchainTokenServiceAddressString, 64))
+        }
+        interchainTokenServiceAddress1 = p1;
+        interchainTokenServiceAddress2 = p2;
+
```
milapsheth, line 44:
```suggestion
```

contracts/remote-address-validator/RemoteAddressValidator.sol
```diff
@@ -60,6 +74,16 @@ contract RemoteAddressValidator is IRemoteAddressValidator, Upgradable {
         return s;
     }
 
+    function _interchainTokenServiceAddressString() internal view returns (string memory interchainTokenServiceAddressString) {
+        interchainTokenServiceAddressString = new string(42);
```
milapsheth, line 78: add a comment on how this works for future maintenance

contracts/remote-address-validator/RemoteAddressValidator.sol
```diff
@@ -60,6 +74,16 @@ contract RemoteAddressValidator is IRemoteAddressValidator, Upgradable {
...
+        uint256 p1 = interchainTokenServiceAddress1;
+        uint256 p2 = interchainTokenServiceAddress2;
+        assembly{
```
milapsheth, line 81:
```suggestion

        assembly {
```

contracts/remote-address-validator/RemoteAddressValidator.sol
```diff
@@ -16,8 +16,12 @@ contract RemoteAddressValidator is IRemoteAddressValidator, Upgradable {
     mapping(string => string) public remoteAddresses;
     address public immutable interchainTokenServiceAddress;
     bytes32 public immutable interchainTokenServiceAddressHash;
```
milapsheth, line 18: this is ununsed now?

contracts/interchain-token-service/InterchainTokenService.sol
```diff
@@ -678,11 +676,11 @@ contract InterchainTokenService is
         address tokenAddress = getStandardizedTokenAddress(tokenId);
         address tokenManagerAddress = getTokenManagerAddress(tokenId);
         address distributor = distributorBytes.length > 0 ? distributorBytes.toAddress() : tokenManagerAddress;
```
milapsheth, line 678: token manager is always a distributor now, so don't need this?

contracts/interchain-token-service/InterchainTokenService.sol
```diff
@@ -678,11 +676,11 @@ contract InterchainTokenService is
...
+        if (distributor == address(0)) revert ZeroAddress();
```
milapsheth, line 679: user can pass 0 for the distributor on the origin if they don't want to give mint/burn permission to anyone other than token manager.

Also, `distributor` is being passed as `mintTo` below. StandardizedToken should check if `mintTo` is `0` and skip minting in that case (i think rn it'll fail).

contracts/interchain-token-service/InterchainTokenService.sol
```diff
@@ -678,11 +676,11 @@ contract InterchainTokenService is
...
         _deployStandardizedToken(tokenId, distributor, name, symbol, decimals, 0, distributor);
-        TokenManagerType tokenManagerType = distributor == tokenManagerAddress ? TokenManagerType.MINT_BURN : TokenManagerType.LOCK_UNLOCK;
         _deployTokenManager(
             tokenId,
-            tokenManagerType,
+            TokenManagerType.MINT_BURN,
             abi.encode(operatorBytes.length == 0 ? address(this).toBytes() : operatorBytes, tokenAddress)
```
milapsheth, line 684: can you remove ternary operator usage everywhere? Slither can't analyze contracts relying on that.

contracts/utils/StandardizedTokenDeployer.sol
```diff
@@ -14,21 +14,18 @@ import { StandardizedTokenProxy } from '../proxies/StandardizedTokenProxy.sol';
...
-     * @param implementationLockUnlockAddress_ Address of the StandardizedTokenLockUnlock contract
-     * @param implementationMintBurnAddress_ Address of the StandardizedTokenMintBurn contract
+     * @param implementationAddress_ Address of the StandardizedToken contract
      */
-    constructor(address deployer_, address implementationLockUnlockAddress_, address implementationMintBurnAddress_) {
-        if (deployer_ == address(0) || implementationLockUnlockAddress_ == address(0) || implementationMintBurnAddress_ == address(0))
+    constructor(address deployer_, address implementationAddress_) {
+        if (deployer_ == address(0) || implementationAddress_ == address(0))
             revert AddressZero();
         deployer = Create3Deployer(deployer_);
```
milapsheth, line 27: We should probably add an `ICreate3Deployer` interface in gmp-sdk instead of importing the actual contract. @deanamiel

contracts/interfaces/IInterchainToken.sol
```diff
@@ -2,12 +2,20 @@
...
  */
-interface IInterchainToken is IERC20 {
+interface IInterchainToken {
+
+    /**
+     * @notice Getter for the tokenManager used for this token.
+     * @dev Needs to be overwitten.
+     * @return tokenManager the TokenManager called to facilitate cross chain transfers.
+     */
+    function getTokenManager() external view returns (ITokenManager tokenManager);
```
milapsheth, line 17:
```suggestion
    function tokenManager() external view returns (ITokenManager tokenManager);
```
We're inconsistent with `get*` vs the name itself for getters in different places. The latter seems to be a common practice.

milapsheth, line 17:
```suggestion
    function getTokenManager() external view returns (address tokenManager);
```
removes the interface dep (which itself has a lot of deps) so it's more self contained. a bit more incovenient in the main contract but seems to be an acceptable trade off

milapsheth: A broader question here. `IInterchainToken` is supposed to represent the interchain token standard. This doesn't technically depend on having a token manager. Should this be added to the standardized token interface instead?

contracts/interchain-token/InterchainToken.sol
```diff
@@ -20,17 +20,6 @@ abstract contract InterchainToken is IInterchainToken, ERC20 {
      */
     function getTokenManager() public view virtual returns (ITokenManager tokenManager);
```
milapsheth, line 21: do we need to inherit from ERC20 here anymore vs just implementing IERC20?

contracts/interchain-token-service/InterchainTokenService.sol
Foivos: It allows users to specify an 'empty' distributor to make sure that no one (other than the tokenManager) can mint/burn, regardless of the chain architecture. I could remove it but I do like this potential feature.

contracts/remote-address-validator/RemoteAddressValidator.sol
milapsheth: fine to keep, but maybe add a getter for it then, so it's easily queryable?

contracts/interchain-token-service/InterchainTokenService.sol
milapsheth: hmm, if the bytes are empty, why not just set `distributor` to `0` instead of the token manager. Token manager will already have mint/burn permission separately, so the distributor doesn't need to be set?

I guess a benefit of this is to allow the token manager to change the distributor in the future? And if the user doesn't want to allow that, they can explicitly pass distributor as `address(0)`. Ok, this makes more sense to me now

contracts/token-implementations/StandardizedToken.sol
```diff
@@ -49,11 +56,14 @@ abstract contract StandardizedToken is InterchainToken, ERC20Permit, Implementat
...
+            (tokenManagerAddress, distributor_, tokenName, symbol, decimals) = abi.decode(
+                params,
+                (address, address, string, string, uint8)
+            );
             _setDistributor(distributor_);
-            tokenManager = tokenManager_;
+            tokenManager_ = tokenManagerAddress;
             _setDomainTypeSignatureHash(tokenName);
             name = tokenName;
         }
```
milapsheth, line 69: skip mint below if `mintTo` is `0`

contracts/remote-address-validator/RemoteAddressValidator.sol
Foivos: it is public so it is easily queryable. (interchainTokenServiceAddress() is a function that can be called)

contracts/interchain-token-service/InterchainTokenService.sol
Foivos: done

milapsheth (approved):

PR #97, opened 2023-07-28
fix: flow limit should have a tighter check

milapsheth:


Foivos (approved):

contracts/utils/FlowLimit.sol
```diff
@@ -106,9 +113,11 @@ contract FlowLimit is IFlowLimit {
     function _addFlowOut(uint256 flowOutAmount) internal {
         uint256 flowLimit = getFlowLimit();
         if (flowLimit == 0) return;
```
deanamiel, line 115: Do we want to keep track of flow in/flow out even if the flowLimit is infinite? In case it gets changed to a finite value after and previous flows affect the current allowance? This issue was brought up in code4rena.

contracts/utils/FlowLimit.sol
```diff
@@ -119,9 +128,11 @@ contract FlowLimit is IFlowLimit {
     function _addFlowIn(uint256 flowInAmount) internal {
         uint256 flowLimit = getFlowLimit();
         if (flowLimit == 0) return;
+
```
deanamiel, line 131: Same question here

contracts/utils/FlowLimit.sol
milapsheth: flow limit doesn't need to be applied retroactively, it should be applicable on future transfers. And doesn't make sense to pay the significant gas penalty for tokens with no flow limits anyways.

contracts/utils/FlowLimit.sol
milapsheth: same

deanamiel (approved):

re1ro (approved):

PR #107, opened 2023-09-07
feat: another example

Foivos:


test/tokenServiceFullFlow.js
```diff
@@ -235,4 +236,104 @@ describe('Interchain Token Service', () => {
             await expect(token.burn(newAddress, amount)).to.be.revertedWithCustomError(token, 'NotDistributor');
         });
     });
+
+    describe('Full pre-existing token registration and token send', async () => {
```
milapsheth, line 240: We want to show the following. Deploy two normal mintable burnable tokens, one on each chain (instead of a test interchain token contract). Deploy token managers for each with mint/burn type under the same token id. Transfer ownership of each token to the token manager. Send a transfer in each direction.

re1ro (approved):

milapsheth (approved):

PR #113, opened 2023-10-02
feat: improved test coverage part two

deanamiel:
PR number two for increased test coverage for ITS as per [AXE-1145](https://axelarnetwork.atlassian.net/jira/software/c/projects/AXE/boards/21?modal=detail&selectedIssue=AXE-1145)

[AXE-1145]: https://axelarnetwork.atlassian.net/browse/AXE-1145?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


contracts/interchain-token-service/InterchainTokenService.sol
```diff
@@ -807,7 +807,7 @@ contract InterchainTokenService is
     function _decodeMetadata(bytes memory metadata) internal pure returns (uint32 version, bytes memory data) {
         data = new bytes(metadata.length - 4);
         assembly {
-            version := shr(224, mload(data))
+            version := shr(224, mload(add(metadata, 32)))
```
deanamiel, line 810: Please review this change @milapsheth @Foivos @re1ro 
In testing `interchainTransfer` I found that the version would always be zero regardless of what `metadata` I passed in to the function. In the original code `data` is uninitialized and will always be all zero bytes which means that `version` will always revert back to zero regardless of what `metadata` is passed into the decode function, here is my proposed fix. Let me know your thoughts.

Foivos: this is correct

Foivos (approved):

scripts/deploy.js
```diff
@@ -6,7 +6,7 @@ const { create3DeployContract, getCreate3Address } = require('@axelar-network/ax
 
 async function deployContract(wallet, contractName, args = []) {
     const factory = await ethers.getContractFactory(contractName, wallet);
-    const contract = await factory.deploy(...args);
+    const contract = await factory.deploy(...args).then((d) => d.deployed());
```
milapsheth, line 9:
```suggestion
    const contract = await factory.deploy(...args);
    await contract.deployTransaction.wait(network.config.confirmations);
```

milapsheth (approved):

PR #116, opened 2023-10-03
feat: custom token registrars

Foivos:
[AXE-2000]

Token registrars help guarantee the integrity of token deployments (token decimals, mint amount etc.) to additional chains via ITS in the future

[AXE-2000]: https://axelarnetwork.atlassian.net/browse/AXE-2000?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


re1ro (approved):

re1ro (approved):

re1ro (approved):

contracts/token-registrars/StandardizedTokenRegistrar.sol
```diff
@@ -0,0 +1,159 @@
...
+        uint256 gasValue
+    ) internal {
+        // slither-disable-next-line arbitrary-send-eth
+        service.deployAndRegisterRemoteStandardizedToken{ value: gasValue }(
+            salt,
+            tokenName,
+            tokenSymbol,
+            tokenDecimals,
+            distributor,
+            '',
```
milapsheth, line 152: mintTo isn't being provided so I think the initial mint will be skipped? Can you add that, and add test coverage for it?

contracts/token-registrars/StandardizedTokenRegistrar.sol
```diff
@@ -0,0 +1,159 @@
...
+
+        service.deployAndRegisterStandardizedToken(salt, name, symbol, decimals, mintAmount, distributor);
+        ITokenManager tokenManager = ITokenManager(service.getTokenManagerAddress(tokenId));
+        tokenManager.transferOperatorship(sender);
+
+        IStandardizedToken token = IStandardizedToken(service.getStandardizedTokenAddress(tokenId));
+        token.safeTransfer(sender, mintAmount);
+    }
+
+    function deployRemoteStandarizedToken(
```
milapsheth, line 84: Can we combine the two deploy methods into one? The interchain address tracker has the current chain, so we can specialize the logic by comparing the chain name being passed. Reduces number of entrypoints. Perhaps we can simplify this in ITS itself.

contracts/token-registrars/StandardizedTokenRegistrar.sol
```diff
@@ -0,0 +1,159 @@
...
+        tokenAddress = service.getStandardizedTokenAddress(getStandardizedTokenId(deployer, salt));
+    }
+
+    function deployStandardizedToken(
+        bytes32 salt,
+        string calldata name,
+        string calldata symbol,
+        uint8 decimals,
+        uint256 mintAmount,
+        address distributor
```
milapsheth, line 70: let operator still be different than sender, since it's an argument for the remote deploy method as well?

contracts/token-registrars/StandardizedTokenRegistrar.sol
```diff
@@ -0,0 +1,159 @@
...
+            ITokenManager tokenManager = ITokenManager(service.getTokenManagerAddress(tokenId));
+
+            tokenName = token.name();
+            tokenSymbol = token.symbol();
+            tokenDecimals = token.decimals();
+            if (additionalDistributor != address(0)) {
+                if (!token.isDistributor(additionalDistributor)) revert NotDistributor(additionalDistributor);
+                distributor = additionalDistributor.toBytes();
+            } else if (mintAmount != 0) {
+                revert NonZeroMintAmount();
```
milapsheth, line 113: add a comment that if distributor is not set than mintAmount must be 0 to enforce that additional token deployments can't change it

contracts/token-registrars/StandardizedTokenRegistrar.sol
```diff
@@ -0,0 +1,159 @@
...
+            }
+            if (optionalOperator != address(0)) {
```
milapsheth, line 115: newline

contracts/token-registrars/CanonicalTokenRegistrar.sol
```diff
@@ -0,0 +1,98 @@
...
+        address tokenAddress,
+        string calldata destinationChain,
+        bytes calldata destinationAddress,
+        uint256 amount,
+        uint256 gasValue
+    ) external payable {
+        // This ensures that the token manager has been deployed by this address, so it's safe to trust it.
+        bytes32 tokenId = getCanonicalTokenId(tokenAddress);
+        // slither-disable-next-line unused-return
+        service.getValidTokenManagerAddress(tokenId);
```
milapsheth, line 87: interchainTransfer will fail if it's not valid so is this check needed?

contracts/token-registrars/StandardizedTokenRegistrar.sol
```diff
@@ -0,0 +1,159 @@
...
+
+    bytes32 private constant CONTRACT_ID = keccak256('standardized-token-registrar');
+
+    struct DeployParams {
+        address deployer;
+        bytes distributor;
+        bytes operator;
+    }
+
+    mapping(bytes32 => DeployParams) public deploymentParameterMap;
```
milapsheth, line 34: unused?

contracts/token-registrars/StandardizedTokenRegistrar.sol
```diff
@@ -0,0 +1,159 @@
...
+import { IStandardizedTokenRegistrar } from '../interfaces/IStandardizedTokenRegistrar.sol';
+import { ITokenManagerType } from '../interfaces/ITokenManagerType.sol';
+import { ITokenManager } from '../interfaces/ITokenManager.sol';
+import { IStandardizedToken } from '../interfaces/IStandardizedToken.sol';
+
+import { Multicall } from '../utils/Multicall.sol';
+
+import { AddressBytesUtils } from '../libraries/AddressBytesUtils.sol';
+
+contract StandardizedTokenRegistrar is IStandardizedTokenRegistrar, ITokenManagerType, Multicall, Upgradable {
```
milapsheth, line 18: Can we combine the two registrars into one? Easier to manage operations around it and for UI to integrate

contracts/token-registrars/CanonicalTokenRegistrar.sol
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    function getCanonicalTokenId(address tokenAddress) public view returns (bytes32 tokenId) {
+        tokenId = service.getCustomTokenId(address(this), getCanonicalTokenSalt(tokenAddress));
+    }
+
+    function registerCanonicalToken(address tokenAddress) external payable returns (bytes32 tokenId) {
+        bytes memory params = abi.encode('', tokenAddress);
+        bytes32 salt = getCanonicalTokenSalt(tokenAddress);
+        tokenId = service.deployCustomTokenManager(salt, TokenManagerType.LOCK_UNLOCK, params);
```
milapsheth, line 50: Shall we remove the canonical register/deploy methods from ITS?

contracts/token-registrars/StandardizedTokenRegistrar.sol
milapsheth: Similarly, we could combine deploy custom token manager into one method

contracts/token-registrars/CanonicalTokenRegistrar.sol
```diff
@@ -0,0 +1,98 @@
...
+            '',
+            '',
+            0,
+            '',
+            destinationChain,
+            gasValue
+        );
+    }
+
+    function transferCanonicalToken(
```
milapsheth, line 77: If we can figure out how to make this generic for other token types, then we could drop the initial mint amount from the official ITS deploy payload spec altogether.

Perhaps this approach could work:
- The deploy method for the current chain takes in mintTo/mintAmount argument.
- The deploy remote method doesn't forward mintAmount anymore. It just deploys.
- The UI sets the mintTo address to be the registrar contract.
- The subsequent transfer calls in the multicall then call interchainTransfer method (with an approval to the token manager if necessary).
- This way the user doesn't need to pre-approve a non-existent token.

contracts/token-registrars/StandardizedTokenRegistrar.sol
Foivos: fixed

PR #122, opened 2023-10-09
feat: bump to latest gmp-sdk and cgp

milapsheth:
[AXE-2034]

[AXE-2034]: https://axelarnetwork.atlassian.net/browse/AXE-2034?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


re1ro (approved):

re1ro:
LGTM! That's a big one

test/tokenServiceFullFlow.js
```diff
@@ -56,8 +56,13 @@ describe('Interchain Token Service Full Flow', () => {
         });
 
         it('Should register the token and initiate its deployment on other chains', async () => {
-            const tx1 = await service.populateTransaction.registerCanonicalToken(token.address);
```
re1ro, line 59: @milapsheth the issue is we can't register token and call other commands in the same multicall TX. The [token manager check for code length](https://github.com/axelarnetwork/interchain-token-service/blob/c618c91aa5855639bcfaaa5d7cd60c1a744d8702/contracts/interchain-token-service/InterchainTokenService.sol#L165) fails if in the same TX

Foivos: Are you sure? It's worked before even though the same check was in place.

Foivos (approved):

PR #123, opened 2023-10-09
build: contract sizer [AXE-2035]

blockchainguyy:
Compilation will fail if size exceeds 24kb.


hardhat.config.js
```diff
@@ -2,6 +2,7 @@ require('dotenv').config();
 require('@nomicfoundation/hardhat-toolbox');
 require('solidity-coverage');
 require('solidity-docgen');
+require('hardhat-contract-sizer');
```
milapsheth, line 5: Add a unit test instead? Seems easier to test changes to bring it down if it can at least compile but fail the test

hardhat.config.js
```diff
@@ -44,4 +45,8 @@ module.exports = {
     gasReporter: {
         enabled: process.env.REPORT_GAS !== undefined,
     },
+    contractSizer: {
+        runOnCompile: true,
```
milapsheth, line 49: this shouldn't output on every build. Make this env var dependent as well.

hardhat.config.js
```diff
@@ -44,4 +45,8 @@ module.exports = {
...
+        strict: process.env.FAIL_COMPILE_IF_SIZE_EXCEEDS,
```
milapsheth, line 50:
```suggestion
        strict: process.env.CHECK_CONTRACT_SIZE,
```

hardhat.config.js
re1ro: @blockchainguyy if I am not mistaken when contract size is over the limit, the solidity compiler throws warnings anyway.

milapsheth (approved):

PR #124, opened 2023-10-09
ci: check compilation warnings [AXE-2035]

blockchainguyy:


.github/workflows/test.yaml
```diff
@@ -23,5 +23,21 @@ jobs:
...
+
+            exit 1
+          else
+            exit 0;
+          fi
+
       - name: Test
-        run: npm run test
+        # Running test-without-compile as artifacts already generated in build command
+        run: npm run test-without-compile
```
milapsheth, line 43: This is not needed. It skips compilation automatically if contracts were already compiled

milapsheth (approved):

PR #131, opened 2023-10-24
feat(Service): implementing ValuedExpressExecutable

re1ro:
[AXE-2001]
* [x] `InterchainTokenService`: implementing ValuedExpressExecutable
* [x] `InterchainTokenExecutable`: returning a constant to insure call success
* [x]  `InterchainTokenService`: express events and valued interface from GMP SDK
* [x] Tests: fixed to use new expressExecute method and event signature

[AXE-2001]: https://axelarnetwork.atlassian.net/browse/AXE-2001?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


contracts/interchain-token-service/InterchainTokenService.sol
```diff
@@ -398,42 +400,83 @@ contract InterchainTokenService is
...
     /**
      * @notice Uses the caller's tokens to fullfill a sendCall ahead of time. Use this only if you have detected an outgoing
      * interchainTransfer that matches the parameters passed here.
      * @dev This is not to be used with fee on transfer tokens as it will incur losses for the express caller.
      * @param payload the payload of the receive token
      * @param commandId the sendHash detected at the sourceChain.
      */
-    function expressReceiveToken(bytes calldata payload, bytes32 commandId, string calldata sourceChain) external {
-        if (gateway.isCommandExecuted(commandId)) revert AlreadyExecuted(commandId);
+    function _expressReceiveToken(bytes32 commandId, string calldata sourceChain, bytes calldata payload, bytes32 payloadHash) internal {
```
milapsheth, line 448: Can we just reuse ExpressExecutorTracker contract for storing express info instead of redefining with express caller? We can move the interface to the individual express executables so it doesn't interfere with the inheritance
https://github.com/axelarnetwork/axelar-gmp-sdk-solidity/blob/main/contracts/express/ExpressExecutorTracker.sol#L7

At the least we should use express executor naming convention

contracts/interchain-token-service/InterchainTokenService.sol
```diff
@@ -398,42 +400,83 @@ contract InterchainTokenService is
         );
     }
 
+    // Returns the amount of token that this call is worth. If `tokenAddress` is `0`, then value is in terms of the native token, otherwise it's in terms of the token address.
+    function contractCallValue(
+        string calldata sourceChain,
+        string calldata sourceAddress,
+        bytes calldata payload
+    ) public view virtual onlyRemoteService(sourceChain, sourceAddress) notPaused returns (address, uint256) {
+        (, bytes32 tokenId, , uint256 amount) = abi.decode(payload, (uint256, bytes32, bytes, uint256));
```
milapsheth, line 409: this should have the same checks apply, such as revert on invalid selector since that is not express callable. The relayer will query this method to determine if it can be expressed

contracts/interchain-token-service/InterchainTokenService.sol
```diff
@@ -533,28 +576,68 @@ contract InterchainTokenService is
...
      * @param sourceChain The chain where the transaction originates from
-     * @param sourceAddress The address where the transaction originates from
+     * @param sourceService The address of the remote ITS where the transaction originates from
      * @param payload The encoded data payload for the transaction
      */
     function execute(
         bytes32 commandId,
         string calldata sourceChain,
-        string calldata sourceAddress,
+        string calldata sourceService,
```
milapsheth, line 585: Let's preserve the GMP API naming convention

contracts/interchain-token-service/InterchainTokenService.sol
```diff
@@ -398,42 +400,83 @@ contract InterchainTokenService is
...
+        ITokenManager tokenManager = ITokenManager(getValidTokenManagerAddress(tokenId));
+        return (tokenManager.tokenAddress(), amount);
+    }
+
+    function expressExecute(
+        bytes32 commandId,
+        string calldata sourceChain,
+        string calldata sourceService,
+        bytes calldata payload
+    ) external payable notPaused {
```
milapsheth, line 420: we can keep the paused check for now since this can also do arbitrary contract execution

contracts/interchain-token-service/InterchainTokenService.sol
milapsheth, line 448:
```suggestion
    function _expressExecute(bytes32 commandId, string calldata sourceChain, bytes calldata payload, bytes32 payloadHash) internal {
```
for more consistency

contracts/interfaces/IInterchainTokenService.sol
```diff
@@ -38,14 +45,19 @@ interface IInterchainTokenService is ITokenManagerType, IExpressCallHandler, IAx
         address indexed sourceAddress,
         bytes data
     );
-    event TokenReceived(bytes32 indexed tokenId, string sourceChain, address indexed destinationAddress, uint256 indexed amount);
-    event TokenReceivedWithData(
+    event TokenReceived(
         bytes32 indexed tokenId,
         string sourceChain,
+        bytes sourceAddress,
```
milapsheth, line 51: :+1: Some more gas but good to have for monitoring

contracts/utils/ExpressCallHandler.sol
```diff
@@ -10,27 +10,39 @@ import { IExpressCallHandler } from '../interfaces/IExpressCallHandler.sol';
  * token transfers and token transfers with contract calls between chains. Implements the IExpressCallHandler interface.
  */
 contract ExpressCallHandler is IExpressCallHandler {
```
milapsheth, line 12: Let's use ExpressExecutorTracker naming convention if we don't just reuse that contract.

contracts/utils/ExpressCallHandler.sol
```diff
@@ -10,27 +10,39 @@ import { IExpressCallHandler } from '../interfaces/IExpressCallHandler.sol';
...
     /**
      * @notice Calculates the unique slot for a given express token transfer.
-     * @param payload the payload of the receive token
      * @param commandId The unique hash for this token transfer
+     * @param payloadHash The payload keccak for the receive token
      * @return slot The calculated slot for this token transfer
      */
-    function _getExpressReceiveTokenSlot(bytes calldata payload, bytes32 commandId) internal pure returns (uint256 slot) {
-        slot = uint256(keccak256(abi.encode(PREFIX_EXPRESS_RECEIVE_TOKEN, payload, commandId)));
+    function _getExpressCallerSlot(bytes32 commandId, bytes32 payloadHash) internal pure returns (uint256 slot) {
```
milapsheth, line 22: I think this should at least include sourceChain, otherwise you can make two GMP calls from chains A and B with same payload, express execute using commandId for call A but source chain B, which will emit an express executed event including source chain B, and when GMP call from A arrives it'll still go through but source chain will actually be A. It's a weird behaviour, so I prefer to avoid skipping any of the GMP params here similar to express executor tracker.

sergeynog:
We said we'd change to use "intent" vs express going forward.

milapsheth:
> We said we'd change to use "intent" vs express going forward.

Need to refactor the original contract first, so leaving the name change for a separate PR to not block the main fix

contracts/interchain-token-service/InterchainTokenService.sol
Foivos: Specifically deploying tokens should NOT be expressable. Only receive token and receive token with data should.

contracts/interchain-token-service/InterchainTokenService.sol
```diff
@@ -398,42 +400,83 @@ contract InterchainTokenService is
...
      * @notice Uses the caller's tokens to fullfill a sendCall ahead of time. Use this only if you have detected an outgoing
      * interchainTransfer that matches the parameters passed here.
      * @dev This is not to be used with fee on transfer tokens as it will incur losses for the express caller.
      * @param payload the payload of the receive token
      * @param commandId the sendHash detected at the sourceChain.
      */
-    function expressReceiveToken(bytes calldata payload, bytes32 commandId, string calldata sourceChain) external {
-        if (gateway.isCommandExecuted(commandId)) revert AlreadyExecuted(commandId);
+    function _expressReceiveToken(bytes32 commandId, string calldata sourceChain, bytes calldata payload, bytes32 payloadHash) internal {
+        if (gateway.isCommandExecuted(commandId)) revert AlreadyExecuted();
```
Foivos, line 449: I like having arguments in errors for clarity, why remove them?

contracts/interchain-token-service/InterchainTokenService.sol
```diff
@@ -398,42 +400,83 @@ contract InterchainTokenService is
...
 
-        address caller = msg.sender;
```
Foivos, line 411: This saves some gas if there are no stack too deep issues.

contracts/interchain-token-service/InterchainTokenService.sol
Foivos: I agree

contracts/interchain-token-service/InterchainTokenService.sol
```diff
@@ -533,28 +576,68 @@ contract InterchainTokenService is
...
+        bytes32 /*commandId*/,
+        string calldata /*sourceChain*/,
+        string calldata /*sourceAddress*/,
+        bytes calldata /*payload*/,
+        string calldata /*tokenSymbol*/,
+        uint256 /*amount*/
+    ) external payable {
+        revert ExecuteWithTokenNotSupported();
+    }
+
```
Foivos, line 640: We could remove these if we get contract size issues and they will still revert without an error code, but we can keep them in if we need to for clarity until then.

contracts/interchain-token-service/InterchainTokenService.sol
```diff
@@ -878,7 +964,7 @@ contract InterchainTokenService is
             // slither-disable-next-line reentrancy-events
             emit TokenSent(tokenId, destinationChain, destinationAddress, amount);
 
-            payload = abi.encode(SELECTOR_RECEIVE_TOKEN, tokenId, destinationAddress, amount);
+            payload = abi.encode(SELECTOR_RECEIVE_TOKEN, tokenId, sourceAddress.toBytes(), destinationAddress, amount);
```
Foivos, line 967: Why are you adding the source address here? It seems it is only to emit an event when receiving, and I'd rather save the non-trivial amount of gas instead.

re1ro: I mean without it on the destination chain you don't even know who sent the token? Token transfers shouldn't be anon right?

contracts/interchain-token-service/InterchainTokenService.sol
re1ro: We need them to satisfy the Express interface from GMP SDK

contracts/interchain-token-service/InterchainTokenService.sol
re1ro: It was actually implemented in the `expressExecute` method

contracts/interchain-token-service/InterchainTokenService.sol
re1ro: Having issues with contract size:
```
Transaction reverted: trying to deploy a contract whose code is too large
```

contracts/utils/ExpressCallHandler.sol
re1ro: Reused

contracts/interfaces/IExpressCallHandler.sol
```diff
@@ -5,15 +5,4 @@ pragma solidity ^0.8.0;
 interface IExpressCallHandler {
     error AlreadyExpressCalled();
     error SameDestinationAsCaller();
```
milapsheth, line 7: is this interface needed?

contracts/interchain-token-service/InterchainTokenService.sol
re1ro: Resolved with setting lower optimizer number

contracts/interchain-token-service/InterchainTokenService.sol
```diff
@@ -547,14 +581,43 @@ contract InterchainTokenService is
         if (!gateway.validateContractCall(commandId, sourceChain, sourceAddress, payloadHash)) revert NotApprovedByGateway();
 
         uint256 selector = abi.decode(payload, (uint256));
-        if (selector == SELECTOR_RECEIVE_TOKEN || selector == SELECTOR_RECEIVE_TOKEN_WITH_DATA)
-            return _processReceiveTokenPayload(commandId, sourceChain, payload, selector);
+        if (selector == SELECTOR_RECEIVE_TOKEN || selector == SELECTOR_RECEIVE_TOKEN_WITH_DATA) {
+            address expressCaller = _popExpressExecutor(commandId, sourceChain, sourceAddress, payloadHash);
```
milapsheth, line 585:
```suggestion
            address expressExecutor = _popExpressExecutor(commandId, sourceChain, sourceAddress, payloadHash);
```

contracts/interchain-token-service/InterchainTokenService.sol
```diff
@@ -547,14 +581,43 @@ contract InterchainTokenService is
...
+            _processReceiveTokenPayload(expressCaller, sourceChain, payload, selector);
+
+            if (expressCaller != address(0))
+                emit ExpressExecutionFulfilled(commandId, sourceChain, sourceAddress, payloadHash, expressCaller);
```
milapsheth, line 589:
```suggestion
            if (expressExecutor != address(0))
                emit ExpressExecutionFulfilled(commandId, sourceChain, sourceAddress, payloadHash, expressExecutor);
```

contracts/interchain-token-service/InterchainTokenService.sol
milapsheth: yeah, but we want to fail here as well so the relayer knows before they even try doing expressExecute

contracts/interchain-token-service/InterchainTokenService.sol
```diff
@@ -572,49 +635,52 @@ contract InterchainTokenService is
      * @param payload The encoded data payload to be processed
      */
     function _processReceiveTokenPayload(
-        bytes32 commandId,
+        address expressCaller,
```
milapsheth, line 638:
```suggestion
        address expressExecutor,
```

contracts/interchain-token-service/InterchainTokenService.sol
```diff
@@ -572,49 +635,52 @@ contract InterchainTokenService is
...
-            address expressCaller = _popExpressReceiveToken(payload, commandId);
-            if (expressCaller != address(0)) {
-                amount = tokenManager.giveToken(expressCaller, amount);
-                return;
-            }
+
+        // Return token to the existing express caller
+        if (expressCaller != address(0)) {
+            // slither-disable-next-line unused-return
+            tokenManager.giveToken(expressCaller, amount);
```
milapsheth, line 658:
```suggestion
        if (expressExecutor != address(0)) {
            // slither-disable-next-line unused-return
            tokenManager.giveToken(expressExecutor, amount);
```

contracts/interfaces/IExpressCallHandler.sol
re1ro: Yeah should be deleted

milapsheth (approved):

contracts/interchain-token-service/InterchainTokenService.sol
re1ro: Added

PR #133, opened 2023-10-31
feat: fixed some potential attack for the fee token manager

Foivos:


contracts/token-manager/implementations/TokenManagerLockUnlockFee.sol
```diff
@@ -52,6 +52,11 @@ contract TokenManagerLockUnlockFee is TokenManager, NoReEntrancy, ITokenManagerL
 
         token.safeTransferFrom(from, address(this), amount);
 
+        uint256 diff = token.balanceOf(address(this)) - balanceBefore;
+        if (diff < amount) {
+            amount = diff;
+        }
+
         return token.balanceOf(address(this)) - balanceBefore;
```
milapsheth, line 60:
```suggestion
        return amount;
```

milapsheth (approved):

PR #140, opened 2023-11-02
feat: build abi files into ts encoding files

canhtrinh:


alanrsoares (approved):
LGTM :ship: :it:

alanrsoares (approved):

alanrsoares:
you'll need to use interactive rebase to fix the conventional commit message check

scripts/codegen.ts
```diff
@@ -0,0 +1,171 @@
+#!/usr/bin/env zx
+import { convertCase } from '@axelarjs/utils/case-conversion';
+import { capitalize } from '@axelarjs/utils/string';
```
milapsheth, line 3: Do we need the full utils dep for this? Maybe some lighter dependency for this scenario works instead?

milapsheth (approved):

scripts/codegen.ts
```diff
@@ -0,0 +1,171 @@
...
+
+import fs from 'fs/promises';
```
milapsheth, line 5: Can this file be imported from somewhere? We'd wanna avoid duplication across repos. Maybe we check this into `gmp-sdk` repo?

scripts/codegen.ts
alanrsoares: This is tree-shakeable and won't affect any client

scripts/codegen.ts
alanrsoares: The whole codegen utility?

canhtrinh: If so, the most natural place would be in our evm package in the axelarjs monorepo since other repos aren't ts-native. 

I'm also fine just leaving this here for now as it's an implementation detail in a build script that can be changed later, and won't affect other repos.

.github/workflows/publish-to-npm.yaml
```diff
@@ -1,22 +1,23 @@
 name: 'publish to npm'
 
 on:
-  workflow_dispatch:
-  release:
-    types: [published]
+    workflow_dispatch:
```
re1ro, line 4: @canhtrinh can we keep the indent the same? 2 space is what we use in other yaml files

re1ro (dismissed):

milapsheth (approved):

PR #163, opened 2023-11-07
fix: typescript codegen

alanrsoares:
* fix typescript codegen script
* add ts compilation step


scripts/codegen.ts
```diff
@@ -52,117 +49,138 @@ const getInputType = (input: ABIInputItem) => {
...
-                 * Original abi file: 
-                 * - ${remainingPath}/${pascalName}.json
-                 * 
-                 * DO NOT EDIT MANUALLY
-                 */
-                `;
+/**
+ * Represents the contract folders to be codegen'd under artifacts/contracts
+ */
+const CONTRACT_FOLDERS = ['InterchainTokenService.sol', 'InterchainTokenFactory.sol', 'interchain-token', 'token-manager'];
```
milapsheth, line 55: Can you include everything under contracts folder?

.gitignore
```diff
@@ -115,3 +115,5 @@ artifacts/
 
 info/local.json
 keys.json
+
+typescript/
```
milapsheth, line 119: Missing newline at the end of file

scripts/codegen.ts
alanrsoares: what else needs to go in there? are all of those interactable?

canhtrinh (approved):

scripts/codegen.ts
alanrsoares: there's a lot of stuff that doesn't seem necessary for consumers under the artifacts/contracts folder, like `test`,`proxies`,`interfaces`,`utils`

milapsheth: Utils are relevant for things transferring mint/burn permission. Any reason interfaces should be skipped? Other things can be relevant for custom tokens. I would put in everything except test since it might not just be us using it. What's the package size with that?

milapsheth (approved):

PR #167, opened 2023-11-08
feat: integrate @axelarjs/evm codegen tool

alanrsoares:
changes:
* remove scripts/codegen.ts
* remove `zx`, `tsx`, `viem`  & `package-json` devDependencies
* add `@axelarjs/evm`  to devDependencies
* update package.json scripts

codegen tool usage:

```bash
npx @axelarjs/evm codegen --src artifacts/contracts --out typescript/contracts --exclude test
```
⚠️⚠️⚠️
the cli can receive an optional `--flatten` flag. 
it is intentionally being omitted here to retain folder hierarchy
⚠️⚠️⚠️


package.json
```diff
@@ -33,6 +34,7 @@
   },
   "devDependencies": {
     "@axelar-network/axelar-chains-config": "~0.1.2",
+    "@axelarjs/evm": "^0.1.17",
     "@axelarjs/utils": "^0.1.12",
```
milapsheth, line 38: utils isn't needed anymore, right?

milapsheth (approved):

package.json
alanrsoares: fixed

PR #187, opened 2023-11-22
refactor(Proxy): using generic proxy for ITS and Factory

re1ro:
* [x] `InterchainTokenFactoryProxy`, `InterchainTokenServiceProxy`: replacing with generic `Proxy` from the SDK


contracts/proxies/Proxy.sol
```diff
@@ -0,0 +1,5 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.0;
+
+import { Proxy } from '@axelar-network/axelar-gmp-sdk-solidity/contracts/upgradable/Proxy.sol';
```
milapsheth, line 5: lint warning. maybe add a test contract?

deanamiel (approved):

contracts/proxies/Proxy.sol
re1ro: added an ignore comment

PR #189, opened 2023-11-23
refactor(ITS protocol): unifying interchain transfer payload type

re1ro:
[AXE-2639]
* [x] `InterchainTokenService` protocol: unifying `MESSAGE_TYPE_INTERCHAIN_TRANSFER` with and without data

[AXE-2639]: https://axelarnetwork.atlassian.net/browse/AXE-2639?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/contracts/InterchainTokenService.sol
+++ b/contracts/InterchainTokenService.sol
@@ -403,7 +403,7 @@ contract InterchainTokenService is
      */
     function _expressExecute(bytes32 commandId, string calldata sourceChain, bytes calldata payload) internal {
         (
-            uint256 messageType,
+            ,
             bytes32 tokenId,
             bytes memory sourceAddress,
             bytes memory destinationAddressBytes,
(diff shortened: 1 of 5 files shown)


contracts/InterchainTokenService.sol
```diff
@@ -415,8 +414,20 @@ contract InterchainTokenService is
...
         token.safeTransferFrom(msg.sender, destinationAddress, amount);
 
-        if (messageType == MESSAGE_TYPE_INTERCHAIN_TRANSFER_WITH_DATA) {
-            (, , , , , bytes memory data) = abi.decode(payload, (uint256, bytes32, bytes, bytes, uint256, bytes));
+        if (data.length == 0) {
+            // slither-disable-next-line reentrancy-events
+            emit InterchainTransferReceived(commandId, tokenId, sourceChain, sourceAddress, destinationAddress, amount);
+        } else {
+            // slither-disable-next-line reentrancy-events
+            emit InterchainTransferReceivedWithData(
```
milapsheth, line 422: combine the events for simplicity? Should be easy to test for `keccak256(0)` (if data is empty, could use a precomputed `keccak256(0)` constant)?

re1ro: Let's keep events separate as we have 2 separate events when transfer arrives to the destination and represent different behavior

milapsheth (approved):

PR #206, opened 2023-12-04
test: preserve 0 flow limiter and update full flow test

milapsheth:
- [x] Remove token manager approval comments
- [x] Set `address(0)` as flow limiter in factory to preserve 2 flow limiter invariant


contracts/InterchainTokenFactory.sol
```diff
@@ -153,8 +153,7 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
             token.transferDistributorship(distributor);
 
             tokenManager.removeFlowLimiter(address(this));
-
-            if (distributor != address(0)) tokenManager.addFlowLimiter(distributor);
+            tokenManager.addFlowLimiter(distributor);
```
Foivos, line 156: since there can be any number of flow limiters this does not really matter i don't think.

PR #207, opened 2023-12-04
refactor: rename tests

milapsheth:


contracts/InterchainTokenService.sol
```diff
@@ -899,7 +899,7 @@ contract InterchainTokenService is
      * @return data The data bytes extracted from the metadata.
      */
     function _decodeMetadata(bytes calldata metadata) internal pure returns (uint32 version, bytes memory data) {
-        if (metadata.length < 4) return (version, data);
+        if (metadata.length < 4) return (LATEST_METADATA_VERSION, data);
```
milapsheth, line 902: Default version when not specified should be the latest

PR #214, opened 2023-12-05
refactor(Factory): removing token transfer functions

re1ro:
[AXE-2766]
* [x] `InterchainTokenFactory`: removing token holding and transfer functionality

[AXE-2766]: https://axelarnetwork.atlassian.net/browse/AXE-2766?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/contracts/InterchainTokenFactory.sol
+++ b/contracts/InterchainTokenFactory.sol
@@ -32,7 +32,6 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
     bytes32 private constant CONTRACT_ID = keccak256('interchain-token-factory');
     bytes32 internal constant PREFIX_CANONICAL_TOKEN_SALT = keccak256('canonical-token-salt');
     bytes32 internal constant PREFIX_INTERCHAIN_TOKEN_SALT = keccak256('interchain-token-salt');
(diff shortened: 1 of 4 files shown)


contracts/InterchainTokenFactory.sol
```diff
@@ -147,11 +139,9 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
             IInterchainToken token = IInterchainToken(service.interchainTokenAddress(tokenId));
             ITokenManager tokenManager = ITokenManager(service.tokenManagerAddress(tokenId));
 
-            _setDeployerTokenBalance(tokenId, sender, initialSupply);
-            token.mint(address(this), initialSupply);
+            token.mint(minter != address(0) ? minter : sender, initialSupply);
```
milapsheth, line 142:
```suggestion
            token.mint(sender, initialSupply);
```
`minter` might be a multisig or so, so maybe we send it to `sender` regardless, so they can move it to other chains right away. The sender wallet will already be connected to the portal.

re1ro: Fixed

Foivos (approved):

PR #248, opened 2024-02-02
feat: move modifiers to be included in executeWithToken

Foivos:


contracts/InterchainTokenService.sol
```diff
@@ -868,7 +868,7 @@ contract InterchainTokenService is
         string calldata sourceAddress,
         bytes calldata payload,
         bytes32 payloadHash
-    ) internal {
+    ) internal onlyRemoteService(sourceChain, sourceAddress) whenNotPaused {
```
milapsheth, line 866: also add unit test causing failure on this

milapsheth (approved):

PR #249, opened 2024-02-06
feat: added additional checks for call contract with token

Foivos:
Tests to hit:
for executeWithToken:
- hit `InvalidExpressMessageType(messageType)` on wrong message type.
- hit `CallWithTokenMissmatch` on wrong amount OR token symbol.

for expressExecuteWithToken:
- hit `CallWithTokenMissmatch` on wrong amount OR token symbol.
- `InvalidExpressMessageType(messageType)` on wrong message type. Should be checked already but make sure.


contracts/interfaces/IInterchainTokenService.sol
```diff
@@ -47,6 +47,7 @@ interface IInterchainTokenService is
     error TokenHandlerFailed(bytes data);
     error EmptyData();
     error PostDeployFailed(bytes data);
+    error CallWithTokenMissmatch(bytes payload, string tokenSymbol, uint256 amount);
```
milapsheth, line 50:
```suggestion
    error InvalidGatewayTokenTransfer(bytes payload, address token, string symbol, uint256 amount);
```
perhaps?

contracts/InterchainTokenService.sol
```diff
@@ -675,7 +676,37 @@ contract InterchainTokenService is
...
+
+        if (expressExecutor != address(0)) {
+            emit ExpressExecutionFulfilled(commandId, sourceChain, sourceAddress, payloadHash, expressExecutor);
+        }
+
+        _processInterchainTransferPayload(
+            commandId,
+            expressExecutor,
+            sourceChain,
+            payload
```
milapsheth, line 696: maybe create an internal method for this to avoid duplication?

contracts/InterchainTokenService.sol
```diff
@@ -675,7 +676,37 @@ contract InterchainTokenService is
         if (!gateway.validateContractCallAndMint(commandId, sourceChain, sourceAddress, payloadHash, tokenSymbol, amount))
             revert NotApprovedByGateway();
 
-        _execute(commandId, sourceChain, sourceAddress, payload, payloadHash);
```
milapsheth, line 678: source address === ITS check is being removed again implicitly

contracts/InterchainTokenService.sol
```diff
@@ -655,9 +655,10 @@ contract InterchainTokenService is
         string calldata sourceChain,
```
milapsheth, line 655: update hardhat config to reduce optimizer runs for ITS contract specifically

re1ro (approved):

contracts/utils/Create3Fixed.sol
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * @title Create3 contract
+ * @notice This contract can be used to deploy a contract with a deterministic address that depends only on
+ * the deployer address and deployment salt, not the contract bytecode and constructor parameters.
+ */
+contract Create3Fixed is Create3Address, IDeploy {
+    using ContractAddress for address;
+
+    bytes internal constant CREATE_DEPLOY_BYTECODE =
+        hex'608060405234801561001057600080fd5b50610162806100206000396000f3fe60806040526004361061001d5760003560e01c806277436014610022575b600080fd5b61003561003036600461007b565b610037565b005b8051602082016000f061004957600080fd5b50565b7f4e487b7100000000000000000000000000000000000000000000000000000000600052604160045260246000fd5b60006020828403121561008d57600080fd5b813567ffffffffffffffff808211156100a557600080fd5b818401915084601f8301126100b957600080fd5b8135818111156100cb576100cb61004c565b604051601f8201601f19908116603f011681019083821181831017156100f3576100f361004c565b8160405282815287602084870101111561010c57600080fd5b82602086016020830137600092810160200192909252509594505050505056fea264697066735822122094780ce55d28f1d568f4e0ab1b9dc230b96e952b73d2e06456fbff2289fa27f464736f6c63430008150033';
```
milapsheth, line 19: Create a constant for the keccak256 of this?

hardhat.config.js
```diff
@@ -27,13 +27,37 @@ const optimizerSettings = {
...
+        jumpdestRemover: true,
+        orderLiterals: true,
+        deduplicate: true,
+        cse: process.env.COVERAGE === undefined,
+        constantOptimizer: true,
+        yul: true,
+        yulDetails: {
+            stackAllocation: true,
+        },
+    },
```
milapsheth, line 45:
```suggestion
    ...optimizerSettings,
    runs: 750,  // Reduce runs to keep bytecode size under limit
```
reduce duplication

test/InterchainTokenService.js
```diff
@@ -2712,19 +2709,107 @@ describe('Interchain Token Service', () => {
...
-                'ExecuteWithTokenNotSupported',
+                'Pause',
             );
+
+            await service.setPauseStatus(false).then((tx) => tx.wait);
+        });
+
+        it('Should revert on invalid express message type', async () => {
+            const message = 10;
+            const amount = 100;
```
milapsheth, line 2751: define these earlier to reuse

contracts/utils/Create3Address.sol
```diff
@@ -0,0 +1,29 @@
...
+
+pragma solidity ^0.8.0;
+
+/**
+ * @title Create3Address contract
+ * @notice This contract can be used to predict the deterministic deployment address of a contract deployed with the `CREATE3` technique.
+ */
+contract Create3Address {
+    /// @dev bytecode hash of the CreateDeploy helper contract
+    bytes32 internal immutable createDeployBytecodeHash;
```
milapsheth, line 11: Since we're fixing this value, this doesn't need to be an immutable. Maybe fix the bytecode hash? You can also inherit/import the constant bytecode and it's hash from a separate contract file

contracts/utils/Create3Address.sol
```diff
@@ -0,0 +1,29 @@
...
+
+    constructor(bytes32 createDeployBytecodeHash_) {
```
milapsheth, line 13: use the constant instead

contracts/utils/Create3Address.sol
```diff
@@ -0,0 +1,29 @@
...
+        createDeployBytecodeHash = createDeployBytecodeHash_;
+    }
+
+    /**
+     * @notice Compute the deployed address that will result from the `CREATE3` method.
+     * @param deploySalt A salt to influence the contract address
+     * @return deployed The deterministic contract address if it was deployed
+     */
+    function _create3Address(bytes32 deploySalt) internal view returns (address deployed) {
```
milapsheth, line 22: this can become a pure method after using a constant

contracts/utils/TokenManagerDeployer.sol
```diff
@@ -34,4 +34,8 @@ contract TokenManagerDeployer is ITokenManagerDeployer, Create3 {
 
         if (tokenManager.code.length == 0) revert TokenManagerDeploymentFailed();
     }
+
+    function createDeployHash() external view returns (bytes32 createDeployBytecodeHash_) {
```
milapsheth, line 38: not really needed to be exposed here once it's a constant

test/InterchainTokenService.js
```diff
@@ -2712,19 +2709,107 @@ describe('Interchain Token Service', () => {
...
+
+            const [tokenAddress, returnedAmount] = await service.contractCallWithTokenValue(
+                sourceChain,
+                trustedAddress,
+                payload,
+                symbol,
+                amount,
+            );
+
+            expect(tokenAddress).to.eq(await service.validTokenAddress(tokenId));
```
milapsheth, line 2811: can you add the remaining test coverage as well?

contracts/InterchainTokenFactory.sol
```diff
@@ -60,10 +60,10 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
      * @param chainNameHash_ The hash of the chain name.
      * @param deployer The address of the deployer.
      * @param salt A unique identifier to generate the salt.
-     * @return bytes32 The calculated salt for the interchain token.
+     * @return tokenSalt The calculated salt for the interchain token.
      */
-    function interchainTokenSalt(bytes32 chainNameHash_, address deployer, bytes32 salt) public pure returns (bytes32) {
-        return keccak256(abi.encode(PREFIX_INTERCHAIN_TOKEN_SALT, chainNameHash_, deployer, salt));
+    function interchainTokenSalt(bytes32 chainNameHash_, address deployer, bytes32 salt) public pure returns (bytes32 tokenSalt) {
```
milapsheth, line 65: changed for consistency? can you update the interface as well?

contracts/InterchainTokenService.sol
```diff
@@ -669,13 +670,15 @@ contract InterchainTokenService is
...
         uint256 amount
-    ) external {
-        bytes32 payloadHash = keccak256(payload);
+    ) external onlyRemoteService(sourceChain, sourceAddress) whenNotPaused {
+        _executeWithToken(commandId, sourceChain, sourceAddress, payload, tokenSymbol, amount);
+    }
 
-        if (!gateway.validateContractCallAndMint(commandId, sourceChain, sourceAddress, payloadHash, tokenSymbol, amount))
-            revert NotApprovedByGateway();
+    function _checkPayloadAgainstGatewayData(bytes calldata payload, string calldata tokenSymbol, uint256 amount) internal view {
```
milapsheth, line 677: add a docstring on what this function is checking

contracts/InterchainTokenService.sol
```diff
@@ -882,6 +885,35 @@ contract InterchainTokenService is
...
+        _checkPayloadAgainstGatewayData(payload, tokenSymbol, amount);
+
+        address expressExecutor = _popExpressExecutor(commandId, sourceChain, sourceAddress, payloadHash);
+
+        if (expressExecutor != address(0)) {
+            emit ExpressExecutionFulfilled(commandId, sourceChain, sourceAddress, payloadHash, expressExecutor);
+        }
+
+        if (!gateway.validateContractCallAndMint(commandId, sourceChain, sourceAddress, payloadHash, tokenSymbol, amount))
+            revert NotApprovedByGateway();
```
milapsheth, line 912: move this to the top? It's better to check this first before doing any computation with untrusted data. would be good to add a comment in `executeWithToken` that gateway validation is being done in this method as an optimization to avoid recomputing the payload hash

contracts/InterchainTokenService.sol
```diff
@@ -641,23 +640,25 @@ contract InterchainTokenService is
...
-        uint256 /*amount*/
-    ) public view virtual returns (address, uint256) {
-        revert ExecuteWithTokenNotSupported();
+        string calldata sourceChain,
+        string calldata sourceAddress,
+        bytes calldata payload,
+        string calldata symbol,
+        uint256 amount
+    ) public view virtual onlyRemoteService(sourceChain, sourceAddress) whenNotPaused returns (address, uint256) {
+        _checkPayloadAgainstGatewayData(payload, symbol, amount);
```
milapsheth, line 649: this method assumes that the payload is for a transfer and returns incomprehensible error if decoding fails on a different type. Since `_callContractValue` returns the address and amount, you could check that directly instead of doing this as well. Alternatively, if we want to reuse the method, then just move it below _contractCallValue which checks the payload type first

contracts/interfaces/IInterchainTokenService.sol
```diff
@@ -47,6 +47,7 @@ interface IInterchainTokenService is
     error TokenHandlerFailed(bytes data);
     error EmptyData();
     error PostDeployFailed(bytes data);
+    error InvalidGatewayTokenTransfer(bytes payload, string tokenSymbol, uint256 amount);
```
milapsheth, line 50: Add `tokenId` explicitly maybe?

contracts/utils/Create3Address.sol
```diff
@@ -0,0 +1,27 @@
...
+
+pragma solidity ^0.8.0;
+
+/**
+ * @title Create3Address contract
+ * @notice This contract can be used to predict the deterministic deployment address of a contract deployed with the `CREATE3` technique.
+ */
+contract Create3Address {
+    // slither-disable-next-line too-many-digits
+    bytes internal constant CREATE_DEPLOY_BYTECODE =
```
milapsheth, line 11: add a comment about this being the fixed CreateDeploy contract bytecode being used by ITS, and fixing this allows changes to compiler settings to not affect the address derivation

contracts/utils/Create3Address.sol
```diff
@@ -0,0 +1,27 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.0;
+
+/**
+ * @title Create3Address contract
+ * @notice This contract can be used to predict the deterministic deployment address of a contract deployed with the `CREATE3` technique.
+ */
+contract Create3Address {
```
milapsheth, line 9: Create3AddressFixed for consistency?

contracts/utils/Create3Fixed.sol
```diff
@@ -0,0 +1,45 @@
...
+
+pragma solidity ^0.8.0;
+
+import { IDeploy } from '@axelar-network/axelar-gmp-sdk-solidity/contracts/interfaces/IDeploy.sol';
+import { ContractAddress } from '@axelar-network/axelar-gmp-sdk-solidity/contracts/libs/ContractAddress.sol';
+import { CreateDeploy } from '@axelar-network/axelar-gmp-sdk-solidity/contracts/deploy/CreateDeploy.sol';
+import { Create3Address } from './Create3Address.sol';
+
+/**
+ * @title Create3 contract
```
milapsheth, line 11:
```suggestion
 * @title Create3Fixed contract
```

contracts/utils/Create3Fixed.sol
```diff
@@ -0,0 +1,45 @@
...
+ * @notice This contract can be used to deploy a contract with a deterministic address that depends only on
+ * the deployer address and deployment salt, not the contract bytecode and constructor parameters.
```
milapsheth, line 13: Explain that this uses the fixed CreateDeploy code so compiler settings don't affect address derivation

hardhat.config.js
```diff
@@ -27,13 +27,24 @@ const optimizerSettings = {
...
     settings: {
         evmVersion: process.env.EVM_VERSION || 'london',
         optimizer: optimizerSettings,
     },
 };
+const itsCompilerSettings = {
+    version: '0.8.21',
+    settings: {
+        evmVersion: process.env.EVM_VERSION || 'london',
+        optimizer: itsOptimizerSettings,
```
milapsheth, line 45:
```suggestion
        optimizer: {
            ...optimizerSettings,
            runs: 600, // Reduce runs to keep bytecode size under limit
        },
```
a bit more concise maybe

hardhat.config.js
```diff
@@ -27,13 +27,24 @@ const optimizerSettings = {
         },
     },
 };
+const itsOptimizerSettings = {
+    ...optimizerSettings,
+    runs: 600, // Reduce runs to keep bytecode size under limit
+};
```
milapsheth, line 33:
```suggestion

```

contracts/InterchainTokenService.sol
```diff
@@ -632,7 +626,12 @@ contract InterchainTokenService is
      * @param sourceAddress The address of the remote ITS where the transaction originates from.
      * @param payload The encoded data payload for the transaction.
      */
-    function execute(bytes32 commandId, string calldata sourceChain, string calldata sourceAddress, bytes calldata payload) public {
+    function execute(
+        bytes32 commandId,
+        string calldata sourceChain,
+        string calldata sourceAddress,
+        bytes calldata payload
+    ) public onlyRemoteService(sourceChain, sourceAddress) whenNotPaused {
```
milapsheth, line 634: make the method external?

contracts/InterchainTokenService.sol
```diff
@@ -641,23 +640,25 @@ contract InterchainTokenService is
...
+        string calldata sourceAddress,
+        bytes calldata payload,
+        string calldata symbol,
+        uint256 amount
+    ) public view virtual onlyRemoteService(sourceChain, sourceAddress) whenNotPaused returns (address, uint256) {
+        _checkPayloadAgainstGatewayData(payload, symbol, amount);
+        return _contractCallValue(payload);
     }
 
     function expressExecuteWithToken(
```
milapsheth, line 653: add docstring. also mention that this method checks paused state implicitly when calling `expressExecute`. Also add a unit test to catch this.

test/InterchainTokenService.js
```diff
@@ -1349,6 +1340,87 @@ describe('Interchain Token Service', () => {
...
+            );
+
+            await expectRevert(
+                (gasOptions) => service.executeWithToken(commandId, sourceChain, wallet.address, '0x', tokenSymbol, amount, gasOptions),
+                service,
+                'NotRemoteService',
+            );
+        });
+
+        it('Should revert on execute if the service is paused', async () => {
```
milapsheth, line 1376:
```suggestion
        it('Should revert on executeWithToken if the service is paused', async () => {
```

test/InterchainTokenService.js
```diff
@@ -1349,6 +1340,87 @@ describe('Interchain Token Service', () => {
...
+            await expectRevert(
+                (gasOptions) => service.executeWithToken(commandId, sourceChain, sourceAddress, '0x', tokenSymbol, amount, gasOptions),
+                service,
+                'Pause',
+            );
+
+            await service.setPauseStatus(false).then((tx) => tx.wait);
+        });
+
+        it('Should revert on execute with invalid messageType', async () => {
```
milapsheth, line 1398:
```suggestion
        it('Should revert on executeWithToken with invalid messageType', async () => {
```

test/InterchainTokenService.js
```diff
@@ -1349,6 +1340,87 @@ describe('Interchain Token Service', () => {
...
+        const tokenSymbol = 'TS';
+        const tokenDecimals = 16;
+
+        before(async () => {
+            sourceAddress = service.address;
+            destAddress = wallet.address;
+            await deployFunctions.gateway(tokenName, tokenSymbol, tokenDecimals);
+        });
+
+        it('Should revert on execute if remote address validation fails', async () => {
```
milapsheth, line 1358:
```suggestion
        it('Should revert on executeWithToken if remote address validation fails', async () => {
```

test/InterchainTokenService.js
```diff
@@ -2750,20 +2827,102 @@ describe('Interchain Token Service', () => {
...
+        const decimals = 18;
+        const message = 0;
         const amount = 100;
+        let tokenId;
+
+        before(async () => {
+            [, , tokenId] = await deployFunctions.gateway(name, symbol, decimals);
+        });
+
+        it('Should revert on contractCallValue if not called by remote service', async () => {
```
milapsheth, line 2843: everywhere below
```suggestion
        it('Should revert on contractCallWithTokenValue if not called by remote service', async () => {
```

test/InterchainTokenService.js
```diff
@@ -2750,20 +2827,102 @@ describe('Interchain Token Service', () => {
...
+                trustedAddress,
+                payload,
+                symbol,
+                amount,
+            );
+
+            expect(tokenAddress).to.eq(await service.validTokenAddress(tokenId));
+            expect(returnedAmount).to.eq(amount);
+        });
     });
```
milapsheth, line 2927: add checks for expressExecuteWithToken

contracts/InterchainTokenService.sol
```diff
@@ -669,13 +681,22 @@ contract InterchainTokenService is
...
-    ) external {
-        bytes32 payloadHash = keccak256(payload);
+    ) external onlyRemoteService(sourceChain, sourceAddress) whenNotPaused {
+        _executeWithToken(commandId, sourceChain, sourceAddress, payload, tokenSymbol, amount);
+    }
 
-        if (!gateway.validateContractCallAndMint(commandId, sourceChain, sourceAddress, payloadHash, tokenSymbol, amount))
-            revert NotApprovedByGateway();
+    /**
+     * @notice Check that the tokenId is has a token that is registered in the gateway with the proper tokenSymbol.
```
milapsheth, line 689:
```suggestion
     * @notice Check that the tokenId from the payload is a token that is registered in the gateway with the proper tokenSymbol, with the right amount from the payload.
```

contracts/InterchainTokenService.sol
```diff
@@ -882,6 +898,31 @@ contract InterchainTokenService is
...
+        string calldata sourceAddress,
+        bytes calldata payload,
+        string calldata tokenSymbol,
+        uint256 amount
+    ) internal {
+        bytes32 payloadHash = keccak256(payload);
+
+        address expressExecutor = _getExpressExecutor(commandId, sourceChain, sourceAddress, payloadHash);
+
+        if (!gateway.validateContractCallAndMint(commandId, sourceChain, sourceAddress, payloadHash, tokenSymbol, amount))
```
milapsheth, line 913: check before getter

test/UtilsTest.js
```diff
@@ -306,3 +308,73 @@ describe('InterchainTokenDeployer', () => {
...
+            const bytecode = factory.getDeployTransaction(name, symbol, decimals).data;
+
+            await expect(deployer.deploy(bytecode, salt)).to.emit(deployer, 'Deployed').withArgs(address);
+        });
+
+        if (isHardhat) {
+            it('should deploy to the predicted address with a know salt', async () => {
+                const salt = '0x4943fe1231449cc1baa660716a0cb38ff09af0b2c9acb63d40d9a7ba06d33d21';
+
+                const address = '0x03C2D7E8Fbcc46C62B3DCBB72121818334af2565';
```
milapsheth, line 355: ah, the factory itself needs a constant address for this test to be meaningful. makes more sense to manually derive create2 address here based on the factory's address to fix the derivation so it's not brittle or remove this test for now

test/InterchainTokenService.js
```diff
@@ -1349,6 +1339,90 @@ describe('Interchain Token Service', () => {
...
+        const tokenSymbol = 'TS';
+        const tokenDecimals = 16;
+
+        before(async () => {
+            sourceAddress = service.address;
+            destAddress = wallet.address;
+            await deployFunctions.gateway(tokenName, tokenSymbol, tokenDecimals);
+        });
+
+        it('Should revert on execute with token if remote address validation fails', async () => {
```
milapsheth, line 1357:
```suggestion
        it('Should revert on executeWithToken with token if remote address validation fails', async () => {
```

contracts/InterchainTokenService.sol
Foivos: slither does not like doing it that way, i could ignore it. (it's because we emit an event at the getter, and it wants the event to be emitted before everything else.

contracts/InterchainTokenService.sol
```diff
@@ -626,32 +621,59 @@ contract InterchainTokenService is
      * @param sourceAddress The address of the remote ITS where the transaction originates from.
      * @param payload The encoded data payload for the transaction.
      */
-    function execute(bytes32 commandId, string calldata sourceChain, string calldata sourceAddress, bytes calldata payload) public {
+    function execute(
+        bytes32 commandId,
+        string calldata sourceChain,
+        string calldata sourceAddress,
+        bytes calldata payload
+    ) external onlyRemoteService(sourceChain, sourceAddress) whenNotPaused {
```
re1ro, line 629: Hey I think we don't need that one as it's already modified with `onlyRemoteService` in `_execute`. But does it make sense to add `onlyRemoteService` to the `expressExecute`? So random contracts can't send express TXs?
cc @milapsheth

re1ro (approved):
LGTM! Only question was if we should put Express methods behind the access control modifier

contracts/InterchainTokenService.sol
milapsheth: `contractCallValue` checks it so our relayer shouldn't express (and it whitelists source address as well), so don't need to pay for the extra gas

contracts/InterchainTokenService.sol
milapsheth: can ignore the slither warning

milapsheth (approved):

PR #252, opened 2024-03-07
feat: added a check to prevent zero amount transfers

Foivos:


contracts/InterchainTokenService.sol
```diff
@@ -1048,6 +1048,7 @@ contract InterchainTokenService is
         string memory symbol,
         uint256 gasValue
     ) internal {
+        if (amount == 0) revert ZeroAmount();
```
milapsheth, line 1051:
```suggestion
        if (amount == 0) revert ZeroAmount();

```

milapsheth, line 1051: add test coverage from different interchainTransfer methods

re1ro: ⬆️

milapsheth (approved):

PR #262, opened 2024-04-15
feat: reduce codesize

Foivos:
move more code to token-handler


re1ro (approved):

re1ro:
@Foivos this would also require deploy/upgrade scripts updates right?

PR #266, opened 2024-07-26
feat: try some different ways to reduce its code size

Foivos:


milapsheth:
closing the PR, can be used for reference in the future

PR #275, opened 2024-09-03
chore: incorrect comments updates

ahramy:
[AXE-4727](https://axelarnetwork.atlassian.net/browse/AXE-4727) : incorrect comments update

[AXE-4727]: https://axelarnetwork.atlassian.net/browse/AXE-4727?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/contracts/InterchainTokenFactory.sol
+++ b/contracts/InterchainTokenFactory.sol
@@ -120,8 +120,9 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
      * @param name The name of the token.
      * @param symbol The symbol of the token.
      * @param decimals The number of decimals for the token.
-     * @param initialSupply The amount of tokens to mint initially (can be zero).
-     * @param minter The address to receive the mintership and operatorship of the token.
+     * @param initialSupply The amount of tokens to mint initially (can be zero), allocated to the msg.sender.
+     * @param minter The address to receive the minter and operator role of the token, in addition to ITS. If it is set to `address(0)`,
+     * the additional minter isn't set, and can't be added later. This allows creating tokens that are managed only by ITS, reducing trust assumptions.
      * @return tokenId The tokenId corresponding to the deployed InterchainToken.
      */
     function deployInterchainToken(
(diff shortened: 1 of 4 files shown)


contracts/InterchainTokenFactory.sol
```diff
@@ -121,7 +121,7 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
      * @param symbol The symbol of the token.
      * @param decimals The number of decimals for the token.
      * @param initialSupply The amount of tokens to mint initially (can be zero).
-     * @param minter The address to receive the initially minted tokens.
+     * @param minter The address to receive the mintership and operatorship of the token.
```
milapsheth, line 124:
```suggestion
     * @param minter The address to receive the minter and operator role of the token, in addition to ITS. If it is set to `address(0)`, the additional minter isn't set, and can't be added later. This allows creating tokens that are managed only by ITS, reducing trust assumptions.
```

contracts/InterchainTokenFactory.sol
```diff
@@ -121,7 +121,7 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
      * @param symbol The symbol of the token.
      * @param decimals The number of decimals for the token.
      * @param initialSupply The amount of tokens to mint initially (can be zero).
```
milapsheth, line 123: Mention that initial supply goes to the sender of this method (and not the minter address).

contracts/InterchainTokenService.sol
```diff
@@ -313,8 +313,7 @@ contract InterchainTokenService is
     /**
      * @notice Used to deploy an interchain token alongside a TokenManager in another chain.
      * @dev At least the `gasValue` amount of native token must be passed to the function call. `gasValue` exists because this function can be
-     * part of a multicall involving multiple functions that could make remote contract calls. If the `minter` parameter is empty bytes then
-     * a mint/burn TokenManager is used, otherwise a lock/unlock TokenManager is used.
+     * part of a multicall involving multiple functions that could make remote contract calls.
```
milapsheth, line 316: Explain how the minter is used though. If minter is empty bytes, no additional minter is set on the token, only ITS is allowed to mint. If the token is being deployed on the current chain, `minter` should correspond to an EVM address (as bytes). Otherwise, an encoding appropriate to the destination chain should be used.

contracts/InterchainTokenFactory.sol
ahramy: updated

contracts/InterchainTokenService.sol
ahramy: updated

milapsheth (approved):

PR #276, opened 2024-09-07
fix: add minter check

ahramy:
[AXE-4774](https://axelarnetwork.atlassian.net/browse/AXE-4774)
- Require to add check that the minter isn’t equal to ITS.
- Added test coverage.

[AXE-4774]: https://axelarnetwork.atlassian.net/browse/AXE-4774?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/contracts/InterchainTokenFactory.sol
+++ b/contracts/InterchainTokenFactory.sol
@@ -140,7 +140,6 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
         if (initialSupply > 0) {
             minterBytes = address(this).toBytes();
         } else if (minter != address(0)) {
-            if (minter == address(interchainTokenService)) revert InvalidMinter(minter);
             minterBytes = minter.toBytes();
         }
 
--- a/contracts/InterchainTokenFactory.sol
+++ b/contracts/InterchainTokenFactory.sol
@@ -140,6 +140,7 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
         if (initialSupply > 0) {
             minterBytes = address(this).toBytes();
         } else if (minter != address(0)) {
+            if(minter == address(interchainTokenService)) revert InvalidMinter(minter);
             minterBytes = minter.toBytes();
         }
 
@@ -202,6 +203,7 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
 
             if (minter != address(0)) {
                 if (!token.isMinter(minter)) revert NotMinter(minter);
+                if(minter == address(interchainTokenService)) revert InvalidMinter(minter);
 
                 minter_ = minter.toBytes();
             }
--- a/contracts/InterchainTokenFactory.sol
+++ b/contracts/InterchainTokenFactory.sol
@@ -140,6 +140,7 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
         if (initialSupply > 0) {
             minterBytes = address(this).toBytes();
         } else if (minter != address(0)) {
+            if (minter == address(interchainTokenService)) revert InvalidMinter(minter);
             minterBytes = minter.toBytes();
         }
 
--- a/contracts/InterchainTokenFactory.sol
+++ b/contracts/InterchainTokenFactory.sol
@@ -123,7 +123,7 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
      * @param initialSupply The amount of tokens to mint initially (can be zero), allocated to the msg.sender.
      * @param minter The address to receive the minter and operator role of the token, in addition to ITS. If it is set to `address(0)`,
      * the additional minter isn't set, and can't be added later. This allows creating tokens that are managed only by ITS, reducing trust assumptions.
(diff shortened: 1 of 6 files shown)


contracts/InterchainTokenFactory.sol
```diff
@@ -140,6 +140,7 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
         if (initialSupply > 0) {
             minterBytes = address(this).toBytes();
         } else if (minter != address(0)) {
+            if (minter == address(interchainTokenService)) revert InvalidMinter(minter);
```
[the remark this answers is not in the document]
ahramy: got it. done.

contracts/InterchainTokenService.sol
```diff
@@ -1101,6 +1101,7 @@ contract InterchainTokenService is
 
         address minter;
         if (bytes(minterBytes).length != 0) minter = minterBytes.toAddress();
+        if (minter == address(this)) revert InvalidMinter(minter);
```
milapsheth, line 1104: this helper is used when processing received deploy call as well, where it doesn't make sense to revert since it'll be a no-op anyways. Having thought more about this, I think it's fine to not check in ITS. And just keep the one in the factory. ITS deploy is flexible so the user can set minter to whatever anyways, so we only really care about trustless factory deployments.

milapsheth (approved):

contracts/InterchainTokenService.sol
ahramy: Got it. updated to keep the checking only in the factory. also updated docstring in the factory about the minter.

contracts/InterchainTokenFactory.sol
```diff
@@ -123,6 +123,7 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
      * @param initialSupply The amount of tokens to mint initially (can be zero), allocated to the msg.sender.
      * @param minter The address to receive the minter and operator role of the token, in addition to ITS. If it is set to `address(0)`,
      * the additional minter isn't set, and can't be added later. This allows creating tokens that are managed only by ITS, reducing trust assumptions.
+     * Reverts if the minter is interchainTokenService address (invalid).
```
milapsheth, line 126:
```suggestion
     * Reverts if the minter is the ITS address since it's already added as a minter.
```
consistent with how we refer elsewhere. explain why it reverts instead of just saying it's invalid

contracts/InterchainTokenFactory.sol
```diff
@@ -140,6 +141,7 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
         if (initialSupply > 0) {
             minterBytes = address(this).toBytes();
         } else if (minter != address(0)) {
+            if (minter == address(interchainTokenService)) revert InvalidMinter(minter);
```
milapsheth, line 144:
```suggestion
            if (minter == address(interchainTokenService)) revert InvalidMinter(minter);

```

contracts/InterchainTokenFactory.sol
```diff
@@ -165,7 +167,8 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
      * @notice Deploys a remote interchain token on a specified destination chain.
      * @param originalChainName The name of the chain where the token originally exists.
      * @param salt The unique salt for deploying the token.
-     * @param minter The address to distribute the token on the destination chain.
+     * @param minter The address to distribute the token on the destination chain. If the address is zero,
+     * no additional minter is set on the token. Reverts if the minter is not authorized or interchainTokenService address (invalid).
```
milapsheth, line 171: same as above. `is not authorized` -> `does not have mint permission for the token`.

contracts/InterchainTokenFactory.sol
```diff
@@ -165,7 +167,8 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
      * @notice Deploys a remote interchain token on a specified destination chain.
      * @param originalChainName The name of the chain where the token originally exists.
      * @param salt The unique salt for deploying the token.
-     * @param minter The address to distribute the token on the destination chain.
+     * @param minter The address to distribute the token on the destination chain. If the address is zero,
```
milapsheth, line 170:
```suggestion
     * @param minter The address to receive the minter and operator role of the token, in addition to ITS. If the address is `address(0)`,
```

milapsheth (approved):

contracts/InterchainTokenFactory.sol
ahramy: got it. done

contracts/InterchainTokenFactory.sol
ahramy: updated.

contracts/InterchainTokenFactory.sol
ahramy: got it, done.

contracts/InterchainTokenFactory.sol
ahramy: got it, updated.

PR #277, opened 2024-09-10
fix: preventing original chain name & destination chain name are the same

ahramy:
[AXE-4776](https://axelarnetwork.atlassian.net/browse/AXE-4776)

- In `InterchainTokenFactory.deployRemoteInterchainToken`, check if `destinationChain` is the same as `originalChainName` to prevent remote deployment requests to the original chain.
- Check originalChainName and destinationChain in `InterchainTokenFactory.deployRemoteCanonicalInterchainToken` to ensure that a remote deploy cannot be requested to the original chain.
- Unit test coverage is 100% for `InterchainTokenFactory.sol` and `InterchainTokenService.sol`
```
----------------------------------------|----------|----------|----------|----------|----------------|
File                                    |  % Stmts | % Branch |  % Funcs |  % Lines |Uncovered Lines |
----------------------------------------|----------|----------|----------|----------|----------------|
  InterchainTokenFactory.sol            |      100 |      100 |      100 |      100 |                |
  InterchainTokenService.sol            |      100 |      100 |      100 |      100 |                |

```

[AXE-4776]: https://axelarnetwork.atlassian.net/browse/AXE-4776?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/contracts/InterchainTokenFactory.sol
+++ b/contracts/InterchainTokenFactory.sol
@@ -123,6 +123,7 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
      * @param initialSupply The amount of tokens to mint initially (can be zero), allocated to the msg.sender.
      * @param minter The address to receive the minter and operator role of the token, in addition to ITS. If it is set to `address(0)`,
      * the additional minter isn't set, and can't be added later. This allows creating tokens that are managed only by ITS, reducing trust assumptions.
+     * Reverts if the minter is the ITS address since it's already added as a minter.
      * @return tokenId The tokenId corresponding to the deployed InterchainToken.
      */
     function deployInterchainToken(
@@ -140,6 +141,8 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
         if (initialSupply > 0) {
             minterBytes = address(this).toBytes();
         } else if (minter != address(0)) {
+            if (minter == address(interchainTokenService)) revert InvalidMinter(minter);
+
             minterBytes = minter.toBytes();
         }
 
@@ -165,7 +168,8 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
      * @notice Deploys a remote interchain token on a specified destination chain.
      * @param originalChainName The name of the chain where the token originally exists.
      * @param salt The unique salt for deploying the token.
-     * @param minter The address to distribute the token on the destination chain.
+     * @param minter The address to receive the minter and operator role of the token, in addition to ITS. If the address is `address(0)`,
+     * no additional minter is set on the token. Reverts if the minter does not have mint permission for the token.
      * @param destinationChain The name of the destination chain.
      * @param gasValue The amount of gas to send for the deployment.
      * @return tokenId The tokenId corresponding to the deployed InterchainToken.
@@ -202,6 +206,7 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
 
             if (minter != address(0)) {
                 if (!token.isMinter(minter)) revert NotMinter(minter);
+                if (minter == address(interchainTokenService)) revert InvalidMinter(minter);
 
                 minter_ = minter.toBytes();
             }
--- a/contracts/InterchainTokenFactory.sol
+++ b/contracts/InterchainTokenFactory.sol
@@ -201,15 +201,8 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
         string memory destinationChain,
         uint256 gasValue
     ) external payable returns (bytes32 tokenId) {
-        bytes32 chainNameHash_;
-
-        if (bytes(originalChainName).length == 0) {
-            chainNameHash_ = chainNameHash;
-        } else {
-            revert NotSupported();
-        }
-
-        tokenId = _deployRemoteInterchainToken(chainNameHash_, salt, minter, destinationChain, gasValue);
+        if (bytes(originalChainName).length != 0) revert NotSupported();
+        tokenId = _deployRemoteInterchainToken(chainNameHash, salt, minter, destinationChain, gasValue);
     }
 
     /**
@@ -333,15 +326,8 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
         string calldata destinationChain,
         uint256 gasValue
     ) external payable returns (bytes32 tokenId) {
-        bytes32 chainNameHash_;
-
-        if (bytes(originalChain).length == 0) {
-            chainNameHash_ = chainNameHash;
-        } else {
-            revert NotSupported();
-        }
-
-        tokenId = _deployRemoteCanonicalInterchainToken(chainNameHash_, originalTokenAddress, destinationChain, gasValue);
+        if (bytes(originalChain).length != 0) revert NotSupported();
+        tokenId = _deployRemoteCanonicalInterchainToken(chainNameHash, originalTokenAddress, destinationChain, gasValue);
     }
 
     /**
--- a/contracts/InterchainTokenFactory.sol
+++ b/contracts/InterchainTokenFactory.sol
@@ -191,11 +191,54 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
             if (bytes(originalChainName).length == 0) {
                 chainNameHash_ = chainNameHash;
             } else {
-                revert NotSupported();
+                chainNameHash_ = keccak256(bytes(originalChainName));
             }
 
             address sender = msg.sender;
             salt = interchainTokenSalt(chainNameHash_, sender, salt);
+            tokenId = interchainTokenService.interchainTokenId(TOKEN_FACTORY_DEPLOYER, salt);
+
+            IInterchainToken token = IInterchainToken(interchainTokenService.interchainTokenAddress(tokenId));
+
+            tokenName = token.name();
+            tokenSymbol = token.symbol();
+            tokenDecimals = token.decimals();
+
+            if (minter != address(0)) {
+                if (!token.isMinter(minter)) revert NotMinter(minter);
+                if (minter == address(interchainTokenService)) revert InvalidMinter(minter);
+
+                minter_ = minter.toBytes();
+            }
+        }
+
+        tokenId = _deployInterchainToken(salt, destinationChain, tokenName, tokenSymbol, tokenDecimals, minter_, gasValue);
+    }
+
+    /**
+     * @notice The deployRemoteInterchainToken with originalChainName allows deploying a remote interchain token back to the same chain, which could cause issues with ITS Hub balance tracking.
+     * To fix this, the deployRemoteInterchainToken function was overloaded to prevent self-deployemnt while still allowing self-transfers.
+     * @param salt The unique salt for deploying the token.
+     * @param minter The address to receive the minter and operator role of the token, in addition to ITS. If the address is `address(0)`,
+     * no additional minter is set on the token. Reverts if the minter does not have mint permission for the token.
+     * @param destinationChain The name of the destination chain.
+     * @param gasValue The amount of gas to send for the deployment.
+     * @return tokenId The tokenId corresponding to the deployed InterchainToken.
+     */
+    function deployRemoteInterchainToken(
+        bytes32 salt,
+        address minter,
+        string memory destinationChain,
+        uint256 gasValue
+    ) external payable returns (bytes32 tokenId) {
+        string memory tokenName;
+        string memory tokenSymbol;
+        uint8 tokenDecimals;
+        bytes memory minter_ = new bytes(0);
+
+        {
+            address sender = msg.sender;
+            salt = interchainTokenSalt(chainNameHash, sender, salt);
             tokenId = interchainTokenService.interchainTokenId(TOKEN_FACTORY_DEPLOYER, salt);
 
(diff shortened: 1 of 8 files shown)


contracts/InterchainTokenFactory.sol
```diff
@@ -285,6 +286,8 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
             } else {
                 chainNameHash_ = keccak256(bytes(originalChain));
             }
+            if (chainNameHash_ == keccak256(bytes(destinationChain))) revert CannotDeployRemotely(destinationChain);
```
milapsheth, line 289: I think we want to remove the ability to deploy a token with a different `originalChain` instead. And perform this check inside ITS `deploy*` methods instead so you can never submit a self deploy call

Foivos (approved):

contracts/InterchainTokenFactory.sol
ahramy: Had offline discussion. updated.

contracts/InterchainTokenFactory.sol
```diff
@@ -191,7 +191,7 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
             if (bytes(originalChainName).length == 0) {
                 chainNameHash_ = chainNameHash;
             } else {
-                chainNameHash_ = keccak256(bytes(originalChainName));
+                revert NotSupported();
```
milapsheth, line 194: behaviour should be documented

contracts/InterchainTokenFactory.sol
```diff
@@ -283,8 +283,9 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
             if (bytes(originalChain).length == 0) {
                 chainNameHash_ = chainNameHash;
             } else {
-                chainNameHash_ = keccak256(bytes(originalChain));
+                revert NotSupported();
```
milapsheth, line 286: same

contracts/interfaces/IInterchainTokenService.sol
```diff
@@ -48,6 +48,7 @@ interface IInterchainTokenService is
     error PostDeployFailed(bytes data);
     error ZeroAmount();
     error CannotDeploy(TokenManagerType);
+    error CannotDeployRemotely(string chainName);
```
milapsheth, line 51:
```suggestion
    error CannotDeployRemotelyToSelf();
```

contracts/InterchainTokenFactory.sol
milapsheth, line 194: add unit test coverage

contracts/InterchainTokenFactory.sol
milapsheth, line 286: an overloaded version of these methods should be created without the `originalChain` arg

contracts/InterchainTokenFactory.sol
```diff
@@ -215,6 +215,49 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
         tokenId = _deployInterchainToken(salt, destinationChain, tokenName, tokenSymbol, tokenDecimals, minter_, gasValue);
     }
 
+    /**
+     * @notice The deployRemoteInterchainToken with originalChainName allows deploying a remote interchain token back to the same chain, which could cause issues with ITS Hub balance tracking.
+     * To fix this, the deployRemoteInterchainToken function was overloaded to prevent self-deployemnt while still allowing self-transfers.
```
milapsheth, line 220:
```suggestion
     * @notice Deploys a remote interchain token on a specified destination chain.
```
Should document what the method does. This method doesn't need to have any discussions about originalChain. It should be discussed in the other `deployRemoteInterchainToken` (and as a `@dev` comment). Put this method first, so users will see it first and use this by default.

This part of the comment `... which could cause issues with ITS Hub balance tracking ...` is vague and so doesn't make much sense to external users. It's just an internal note, it would be better to mention the larger point, something like: `originalChainName is only allowed to be '', i.e the current chain. Other source chains are not supported anymore to simplify ITS deployment behaviour.`

contracts/InterchainTokenFactory.sol
```diff
@@ -215,6 +215,49 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
...
+        uint256 gasValue
+    ) external payable returns (bytes32 tokenId) {
+        string memory tokenName;
+        string memory tokenSymbol;
+        uint8 tokenDecimals;
+        bytes memory minter_ = new bytes(0);
+
+        {
+            address sender = msg.sender;
+            salt = interchainTokenSalt(chainNameHash, sender, salt);
```
milapsheth, line 241: chainNameHash is not defined

contracts/InterchainTokenFactory.sol
```diff
@@ -215,6 +215,49 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
...
+            tokenId = interchainTokenService.interchainTokenId(TOKEN_FACTORY_DEPLOYER, salt);
```
milapsheth, line 242: avoid duplication. Call this method from the other deployRemoteInterchainToken after performing the originalChainName check

contracts/InterchainTokenFactory.sol
```diff
@@ -299,6 +343,37 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
...
 
+    /**
+     * @notice The deployRemoteCanonicalInterchainToken with originalChain allows deploying a canonical interchain token back to the same chain, which could cause issues with ITS Hub balance tracking.
+     * To fix this, the deployRemoteCanonicalInterchainToken function was overloaded to prevent self-deployemnt while still allowing self-transfers.
+     * @param originalTokenAddress The address of the original token on the original chain.
+     * @param destinationChain The name of the chain where the token will be deployed.
+     * @param gasValue The gas amount to be sent for deployment.
+     * @return tokenId The tokenId corresponding to the deployed InterchainToken.
+     */
+    function deployRemoteCanonicalInterchainToken(
```
milapsheth, line 354: same comments as above

contracts/interfaces/IInterchainTokenFactory.sol
```diff
@@ -131,6 +147,20 @@ interface IInterchainTokenFactory is IUpgradable, IMulticall {
         uint256 gasValue
     ) external payable returns (bytes32 tokenId);
 
+    /**
+     * @notice The deployRemoteCanonicalInterchainToken with originalChain allows deploying a canonical interchain token back to the same chain, which could cause issues with ITS Hub balance tracking.
```
milapsheth, line 151: update docstring and order as per other comments

milapsheth (changes requested):

contracts/InterchainTokenFactory.sol
ahramy: Thanks for comment. updated.

contracts/InterchainTokenFactory.sol
```diff
@@ -175,27 +174,19 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
...
         string memory destinationChain,
         uint256 gasValue
-    ) external payable returns (bytes32 tokenId) {
+    ) public payable returns (bytes32 tokenId) {
         string memory tokenName;
         string memory tokenSymbol;
         uint8 tokenDecimals;
         bytes memory minter_ = new bytes(0);
 
         {
```
milapsheth, line 187: wondering if this additional scoping is still needed now that we removed one variable (maybe stack too deep error goes away)

contracts/InterchainTokenFactory.sol
```diff
@@ -216,8 +207,10 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
     }
 
     /**
-     * @notice The deployRemoteInterchainToken with originalChainName allows deploying a remote interchain token back to the same chain, which could cause issues with ITS Hub balance tracking.
-     * To fix this, the deployRemoteInterchainToken function was overloaded to prevent self-deployemnt while still allowing self-transfers.
+     * @notice Deploys a remote interchain token on a specified destination chain.
```
milapsheth, line 210: Is there a `@deprecated` notion in natspec?
```suggestion
     * @notice Deploys a remote interchain token on a specified destination chain. This method is deprecated and will be removed in the future. Please use the above method instead.
```

contracts/InterchainTokenFactory.sol
```diff
@@ -216,8 +207,10 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
...
+     * @dev originalChainName is only allowed to be '', i.e the current chain.
+     * Other source chains are not supported anymore to simplify ITS deployment behaviour.
```
milapsheth, line 212:
```suggestion
     * Other source chains are not supported anymore to simplify ITS token deployment behaviour.
```

contracts/InterchainTokenFactory.sol
```diff
@@ -226,36 +219,14 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
...
-                if (!token.isMinter(minter)) revert NotMinter(minter);
-                if (minter == address(interchainTokenService)) revert InvalidMinter(minter);
-
-                minter_ = minter.toBytes();
-            }
-        }
-
-        tokenId = _deployInterchainToken(salt, destinationChain, tokenName, tokenSymbol, tokenDecimals, minter_, gasValue);
+        if (bytes(originalChainName).length != 0) revert NotSupported();
+        tokenId = deployRemoteInterchainToken(salt, minter, destinationChain, gasValue);
```
milapsheth, line 229:
```suggestion
        if (bytes(originalChainName).length != 0) revert NotSupported();

        tokenId = deployRemoteInterchainToken(salt, minter, destinationChain, gasValue);
```
separating checks from the logic

contracts/InterchainTokenFactory.sol
```diff
@@ -306,31 +277,22 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
...
-        string calldata originalChain,
         address originalTokenAddress,
         string calldata destinationChain,
         uint256 gasValue
-    ) external payable returns (bytes32 tokenId) {
+    ) public payable returns (bytes32 tokenId) {
         bytes32 salt;
         IInterchainToken token;
 
         {
```
milapsheth, line 293: same comment

contracts/InterchainTokenFactory.sol
```diff
@@ -344,34 +306,23 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
     }
 
     /**
-     * @notice The deployRemoteCanonicalInterchainToken with originalChain allows deploying a canonical interchain token back to the same chain, which could cause issues with ITS Hub balance tracking.
-     * To fix this, the deployRemoteCanonicalInterchainToken function was overloaded to prevent self-deployemnt while still allowing self-transfers.
+     * @notice Deploys a canonical interchain token on a remote chain.
```
milapsheth, line 309: same comments

test/InterchainTokenFactory.js
```diff
@@ -449,14 +446,31 @@ describe('InterchainTokenFactory', () => {
                         },
                     ),
                 tokenFactory,
+                'NotSupported',
+            );
+
+            await expectRevert(
+                (gasOptions) =>
+                    tokenFactory['deployRemoteInterchainToken(string,bytes32,address,string,uint256)'](
+                        '',
```
milapsheth, line 455: add a test to ensure it fails if original Chain is not empty (double check test coverage for the repo, so it doesn't drop when making a PR)

contracts/InterchainTokenFactory.sol
ahramy: That's a great call, actually it works without additional scoping. Updated.

contracts/InterchainTokenFactory.sol
ahramy: There's no `@deprecated` notion in Natspec: https://docs.soliditylang.org/en/latest/natspec-format.html#tags
Updated it as the suggestion.

contracts/InterchainTokenFactory.sol
ahramy: updated.

contracts/InterchainTokenFactory.sol
ahramy: updated.

contracts/InterchainTokenFactory.sol
ahramy: updated.

contracts/InterchainTokenFactory.sol
ahramy: updated.

test/InterchainTokenFactory.js
ahramy: `Fails if original chain is not empty` is tested in line 449 with `chainName == "Test"`.
```
await expectRevert(
                (gasOptions) =>
                    tokenFactory['deployRemoteInterchainToken(string,bytes32,address,string,uint256)'](
                        chainName,
                        salt,
                        otherWallet.address,
                        destinationChain,
                        gasValue,
                        {
                            ...gasOptions,
                            value: gasValue,
                        },
                    ),
                tokenFactory,
                'NotSupported',
            );
```
Test coverage for `InterchainTokenFactory.sol` and `InterchainTokenService.sol ` are all 100%. Is there something I'm missing about the test case?

milapsheth (approved):

PR #278, opened 2024-09-17
fix: modified wrong data decoding in _checkPayloadAgainstGatewayData

ahramy:
[AXE-4792](https://axelarnetwork.atlassian.net/browse/AXE-4792)
- Modified [L-01] Wrong data decoding in _checkPayloadAgainstGatewayData
- Unit test coverage remains the same

[AXE-4792]: https://axelarnetwork.atlassian.net/browse/AXE-4792?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/contracts/InterchainTokenService.sol
+++ b/contracts/InterchainTokenService.sol
@@ -707,7 +707,7 @@ contract InterchainTokenService is
      * @param amount The amount for the call contract with token.
      */
     function _checkPayloadAgainstGatewayData(bytes memory payload, string calldata tokenSymbol, uint256 amount) internal view {
-        (, bytes32 tokenId, , , uint256 amountInPayload) = abi.decode(payload, (uint256, bytes32, uint256, uint256, uint256));
+        (, bytes32 tokenId, , , uint256 amountInPayload) = abi.decode(payload, (uint256, bytes32, bytes, bytes, uint256));
 
         if (validTokenAddress(tokenId) != gateway.tokenAddresses(tokenSymbol) || amount != amountInPayload)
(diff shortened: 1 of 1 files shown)


contracts/InterchainTokenService.sol
```diff
@@ -707,7 +707,7 @@ contract InterchainTokenService is
      * @param amount The amount for the call contract with token.
      */
     function _checkPayloadAgainstGatewayData(bytes memory payload, string calldata tokenSymbol, uint256 amount) internal view {
-        (, bytes32 tokenId, , , uint256 amountInPayload) = abi.decode(payload, (uint256, bytes32, uint256, uint256, uint256));
+        (, bytes32 tokenId, , , uint256 amountInPayload) = abi.decode(payload, (uint256, bytes32, bytes, bytes, uint256));
```
Foivos, line 710: You should test this but I am pretty sure there is no difference between `uint256` and `bytes` when decoding if you ignore them, but using `bytes` costs more.

ahramy: I was able to confirm using `bytes` costs more gas.
```
(, bytes32 tokenId, , , uint256 amountInPayload) = abi.decode(payload, (uint256, bytes32, bytes, bytes, uint256));

vs

(, bytes32 tokenId, , , uint256 amountInPayload) = abi.decode(payload, (uint256, bytes32, uint256, uint256, uint256));
```

Hence, just added a comment here.

milapsheth (approved):

PR #279, opened 2024-09-20
fix: remove redundant check on the v value

ahramy:
[AXE-4803](https://axelarnetwork.atlassian.net/browse/AXE-4803)
- Remove Unnecessary v value check in ERC20Permit#permit()
- The ERC20Permit contract implements an unnecessary check for the signature recovery ID (v) before calling the ecrecover function. This check is redundant as ecrecover inherently validates these values.
- The ecrecover function and the subsequent check for a valid recovered address are sufficient to ensure the signature's validity.
- Addressing it will optimize gas usage and simplify the contract code. This change aligns with best practices in Ethereum smart contract development, emphasizing efficiency and clarity in code implementation.

References:
- https://twitter.com/alexberegszaszi/status/1534461421454606336?s=20&t=H0Dv3ZT2bicx00hLWJk7Fg
- The latest [Openzeppelin's ECDSA](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/530179a71f435e85ae9df9b9f12b5637cf229e5c/contracts/utils/cryptography/ECDSA.sol#L136) has removed the v check. by [PR#3591](https://github.com/OpenZeppelin/openzeppelin-contracts/commit/324eda228c07468a0744fc52677e6cebea5dc5c5#diff-ff09871806bcccfd38e43de481f3e7e2fb92134c58e1a1f97b054e2d0d727458) 

[AXE-4803]: https://axelarnetwork.atlassian.net/browse/AXE-4803?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/contracts/interchain-token/ERC20Permit.sol
+++ b/contracts/interchain-token/ERC20Permit.sol
@@ -15,7 +15,6 @@ import { ERC20 } from './ERC20.sol';
 abstract contract ERC20Permit is IERC20, IERC20Permit, ERC20 {
     error PermitExpired();
     error InvalidS();
-    error InvalidV();
     error InvalidSignature();
 
     /**
@@ -72,8 +71,6 @@ abstract contract ERC20Permit is IERC20, IERC20Permit, ERC20 {
         if (block.timestamp > deadline) revert PermitExpired();
 
         if (uint256(s) > 0x7FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF5D576E7357A4501DDFE92F46681B20A0) revert InvalidS();
-
-        if (v != 27 && v != 28) revert InvalidV();
 
         bytes32 digest = keccak256(
             abi.encodePacked(
--- a/test/ERC20Permit.js
+++ b/test/ERC20Permit.js
@@ -172,13 +172,6 @@ describe('ERC20 Permit', () => {
 
         await expectRevert(
             (gasOptions) =>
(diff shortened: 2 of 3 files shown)


contracts/interchain-token/ERC20Permit.sol
```diff
@@ -73,8 +72,6 @@ abstract contract ERC20Permit is IERC20, IERC20Permit, ERC20 {
 
         if (uint256(s) > 0x7FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF5D576E7357A4501DDFE92F46681B20A0) revert InvalidS();
 
-        if (v != 27 && v != 28) revert InvalidV();
```
milapsheth, line 76: i prefer to be more explicit here. also doesn't seem a big enough reason to change the InterchainToken bytecode (existing ITS tokens already have this check, and new ones won't causing an inconsistency that's not a must have)

ahramy: Had discussion, and agreed that we don't want to change the `InterchainToken bytecode`, hence closing this PR.

ahramy:
Had discussion, and agreed that we don't want to change the `InterchainToken bytecode`, hence closing this PR.

PR #282, opened 2024-09-24
chore: make `_spendAllowance` impl match comment

hydrobeam:
[AXE-4831](https://axelarnetwork.atlassian.net/browse/AXE-4831)
- cleaned up the comment to match impl, as recommended.

[AXE-4831]: https://axelarnetwork.atlassian.net/browse/AXE-4831?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/contracts/interchain-token/InterchainToken.sol
+++ b/contracts/interchain-token/InterchainToken.sol
@@ -130,7 +130,6 @@ contract InterchainToken is InterchainTokenStandard, ERC20, ERC20Permit, Minter,
 
     /**
      * @notice A method to be overwritten that will decrease the allowance of the `spender` from `sender` by `amount`.
-     * @dev Needs to be overwritten. This provides flexibility for the choice of ERC20 implementation used. Must revert if allowance is not sufficient.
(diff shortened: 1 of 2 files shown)


contracts/interchain-token/InterchainToken.sol
```diff
@@ -130,7 +130,6 @@ contract InterchainToken is InterchainTokenStandard, ERC20, ERC20Permit, Minter,
 
     /**
      * @notice A method to be overwritten that will decrease the allowance of the `spender` from `sender` by `amount`.
-     * @dev Needs to be overwritten. This provides flexibility for the choice of ERC20 implementation used. Must revert if allowance is not sufficient.
```
milapsheth, line 133: nice to have but i'd rather not change the token bytecode unless needed to avoid two different token versions in production. So, we can close this PR/task for now

PR #284, opened 2024-09-26
fix(its): validate token manager address from token id

ahramy:
[AXE-4705](https://axelarnetwork.atlassian.net/browse/AXE-4705)
- validate token manager address from token id
- coverage remains the same
```
----------------------------------------|----------|----------|----------|----------|----------------|
File                                    |  % Stmts | % Branch |  % Funcs |  % Lines |Uncovered Lines |
----------------------------------------|----------|----------|----------|----------|----------------|
  InterchainTokenFactory.sol            |      100 |      100 |      100 |      100 |                |
  InterchainTokenService.sol            |      100 |      100 |      100 |      100 |                |
```

[AXE-4705]: https://axelarnetwork.atlassian.net/browse/AXE-4705?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/contracts/InterchainTokenService.sol
+++ b/contracts/InterchainTokenService.sol
@@ -413,9 +413,6 @@ contract InterchainTokenService is
 
         IERC20 token;
         {
-            // slither-disable-next-line unused-return
-            validTokenManagerAddress(tokenId);
-
             (bool success, bytes memory returnData) = tokenHandler.delegatecall(
                 abi.encodeWithSelector(ITokenHandler.transferTokenFrom.selector, tokenId, msg.sender, destinationAddress, amount)
             );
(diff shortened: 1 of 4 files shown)


Foivos:
The reason why those checks were not there is because they cost gas, and the call will revert anyway if the `TokenManager` does not exist. Furthermore, the way you are doing it the tokenManagerAddress is calculated twice, once for the check, and once in the `TokenHandler`. If we think we need the checks for sanity I would add them in the `TokenHandler` after the `tokenManagerAddress` is calculated. This would also avoid codesize bloating the service.

ahramy:
> The reason why those checks were not there is because they cost gas, and the call will revert anyway if the `TokenManager` does not exist. Furthermore, the way you are doing it the tokenManagerAddress is calculated twice, once for the check, and once in the `TokenHandler`. If we think we need the checks for sanity I would add them in the `TokenHandler` after the `tokenManagerAddress` is calculated. This would also avoid codesize bloating the service. If you want to make the check in the service then the `tokenManagerAddress` should be passed to the `TokenHandler` instead of the` tokenId` to avoid recalculating it.

Had offline discussion with Milap as well, I think we want to have a long-term solution for this. i.e. migration. Hence, closing this issue for now.

PR #286, opened 2024-10-10
ci: add updated release process for ITS

apolikamixitos:
# What:

Add a new process to release `Interchain Token Service` packages and contracts.

#### `pre-release` Workflow:

Ensures that the changes introduced are ready to get published, as the name suggests, it's a `pre-release` step that creates a PR that could be double-checked before proceeding with publishing a release.

#### `release` Workflow:

This workflow ensures:

- The new versions of the package `@axelar-network/interchain-token-service` are published to `npm`.

image

- New releases/tags are published on `Github`:
   - Includes the source code of the package.
   - The package-related assets (`contracts` ABIs associated with its corresponding package version).

image

- Artifacts being published to `R2` as a secondary backup source and distribution channel:

image
image

#### `release-snapshot` Workflow:

- It's used mainly for `nightly` || `ad-hoc` build to trigger testing purposes, they aren't supposed to be used for `production` environments.

**Notes:**

Some requisites need to take place before merging the PR:

- Ensure that the proper `vars` and `secrets` are set up in the repository config.


.github/workflows/release-snapshot.yaml
```diff
@@ -0,0 +1,34 @@
...
+      - name: Build the modules
+        run: npm run build && npm run build:ts && npm run flatten && npm run clean:artifacts
+
+      - name: Set Snapshot Version
+        run: |
+          npm version 0.0.0-snapshot.$(git rev-parse --short HEAD) --git-tag-version=false
+
+      - name: Publish Snapshot to NPM
+        run: |
+          npm publish --no-git-checks --tag snapshot --access public
```
milapsheth, line 32: Use npm release for the build and publish to avoid duplication

.github/workflows/test.yaml
```diff
@@ -14,14 +14,10 @@ jobs:
     steps:
       - uses: actions/checkout@v4
```
milapsheth, line 15: should these be updated to limit github token availability after use?

.github/workflows/release.yaml
```diff
@@ -1,83 +1,131 @@
...
+      - name: Prepare cross-steps variables
+        run: |
+          export PACKAGE_NAME=$(jq -r '.name' 'package.json');
+          export PUBLISHED_PACKAGE_VERSION=v$(echo '${{ steps.publish-release.outputs.publishedPackages }}' | jq -r ".[] | select(.name == \"${PACKAGE_NAME}\") | .version")
+          export BASE_ARTIFACTS_SEARCH_DIR="./artifacts"
+          export ARTIFACT_VERSIONED_PACKAGE_NAME="interchain-token-service-assets-${PUBLISHED_PACKAGE_VERSION}"
+          export BASE_ARTIFACTS_VERSIONED_DIR="$(dirname ${BASE_ARTIFACTS_SEARCH_DIR})/${ARTIFACT_VERSIONED_PACKAGE_NAME}" # Regardless of the dir type, relative or absolute
+
+          # ex: @axelarnetwork/interchain-token-service
+          echo "PACKAGE_NAME=${PACKAGE_NAME}" >> $GITHUB_ENV
```
milapsheth, line 52:
```suggestion
          echo "package_name=${PACKAGE_NAME}" >> $GITHUB_OUTPUT
```
use github outputs with lower case names so it's a bit more readable later maybe? too many full cap vars

.github/workflows/release.yaml
```diff
@@ -1,83 +1,131 @@
...
+        env:
+          GITHUB_TOKEN: ${{ secrets.PAT_TOKEN }}
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+      # Prepare the variables that will be used across the different next steps
+      - name: Prepare cross-steps variables
+        run: |
+          export PACKAGE_NAME=$(jq -r '.name' 'package.json');
+          export PUBLISHED_PACKAGE_VERSION=v$(echo '${{ steps.publish-release.outputs.publishedPackages }}' | jq -r ".[] | select(.name == \"${PACKAGE_NAME}\") | .version")
```
milapsheth, line 46:
```suggestion
          export PACKAGE_VERSION=v$(echo '${{ steps.publish-release.outputs.publishedPackages }}' | jq -r ".[] | select(.name == \"${PACKAGE_NAME}\") | .version")
```
keep it more concise

.github/workflows/release.yaml
```diff
@@ -1,83 +1,131 @@
...
+          export BASE_ARTIFACTS_SEARCH_DIR="./artifacts"
+          export ARTIFACT_VERSIONED_PACKAGE_NAME="interchain-token-service-assets-${PUBLISHED_PACKAGE_VERSION}"
+          export BASE_ARTIFACTS_VERSIONED_DIR="$(dirname ${BASE_ARTIFACTS_SEARCH_DIR})/${ARTIFACT_VERSIONED_PACKAGE_NAME}" # Regardless of the dir type, relative or absolute
```
milapsheth, line 49:
```suggestion
          export BASE_ARTIFACTS_DIR="./artifacts"
          export ARTIFACT_NAME="interchain-token-service-assets-${PUBLISHED_PACKAGE_VERSION}"
          export ARTIFACTS_DIR="$(dirname ${BASE_ARTIFACTS_DIR})/${ARTIFACT_NAME}" # Regardless of the dir type, relative or absolute
```
a bit more concise

.github/workflows/release.yaml
```diff
@@ -1,83 +1,131 @@
...
-            git checkout -b $new_release_branch
-            git push -u origin $new_release_branch
-          fi
-
-      - name: Bump version and push tag
-        uses: anothrNick/github-tag-action@1.39.0
+          # Renaming to keep the dir structure in the archive matching the release version defined in `BASE_ARTIFACTS_VERSIONED_DIR`
+          mv ${{ env.BASE_ARTIFACTS_SEARCH_DIR }} ${{ env.BASE_ARTIFACTS_VERSIONED_DIR }}
+
+          find "${{ env.BASE_ARTIFACTS_VERSIONED_DIR }}/contracts" -type f -name "*.json" ! -name "*.dbg.json" -print | zip ${{ env.ARTIFACT_VERSIONED_PACKAGE_NAME }}.zip -@
```
milapsheth, line 72: `npm clean:artifacts` deletes *.dbg.json already

.github/workflows/release.yaml
```diff
@@ -1,83 +1,131 @@
...
-          DRY_RUN: true
-
-      - id: parsed-release
-        uses: booxmedialtd/ws-action-parse-semver@v1
+          persist-credentials: false
+      - name: Setup tools
+        uses: ./.github/actions/setup-tools
+        with:
+          install-nodejs: 'true'
+      # Publishes a release in case the release isn't published
```
milapsheth, line 30:
```suggestion

      # Publishes a release in case the release isn't published
```

.github/workflows/release.yaml
milapsheth: i'm also ok with making the env var names more concise instead

.github/workflows/release.yaml
```diff
@@ -1,83 +1,131 @@
...
-          input_string: ${{ steps.release.outputs.tag }}
+          publish: npm run release
+          createGithubReleases: true
+        env:
+          GITHUB_TOKEN: ${{ secrets.PAT_TOKEN }}
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+      # Prepare the variables that will be used across the different next steps
+      - name: Prepare cross-steps variables
```
milapsheth, line 43: Make this into an action so the release flow can be reused in other repos easily?

.github/workflows/release-snapshot.yaml
milapsheth: create a new release-snapshot npm task to so the release cmds are defined in one place

.github/workflows/release.yaml
milapsheth: ok

.github/workflows/release.yaml
milapsheth: yeah, should avoid duplicating, the npm task should be responsible building artifacts appropriately

.github/workflows/release.yaml
milapsheth: sure

milapsheth (approved):

Foivos (approved):

.github/workflows/create-release-pr.yaml
```diff
@@ -0,0 +1,67 @@
...
+jobs:
+  create-release-pr:
+    name: Create Release PR
+    runs-on: blacksmith-2vcpu-ubuntu-2204
+    steps:
+      - name: Check if a branch is whitelisted and maintained
+        id: branch-check
+        run: |
+          BRANCH_NAME="${GITHUB_REF#refs/heads/}"
+          WHITELISTED_BRANCHES=("main" "release/2.*.*")
```
milapsheth:
```suggestion
          WHITELISTED_BRANCHES=("main" "releases/*")
```
should keep it generic. we might want to release for v1.x as well which is live in production. Don't need to be super restrictive here, and have to maintain this workflow with releases

.github/workflows/release.yaml
```diff
@@ -6,6 +6,9 @@ on:
   pull_request:
     branches:
       - main
+      # It needs to be specific based on the maintenance policy that will be set: (e.g.: `release/1.0.*`)
+      # This ensures that the unmaintained old releases are no longer being updated.
+      - release/**
```
milapsheth:
```suggestion
      - 'releases/**'
```

PR #297, opened 2024-11-06
feat(its-factory): allow minter to be arbitrary bytes for remote token deployment

ahramy:
[AXE-6319](https://axelarnetwork.atlassian.net/browse/AXE-6319)

[AXE-6319]: https://axelarnetwork.atlassian.net/browse/AXE-6319?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/contracts/InterchainTokenFactory.sol
+++ b/contracts/InterchainTokenFactory.sol
@@ -155,23 +155,21 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
     /**
      * @notice Deploys a remote interchain token on a specified destination chain.
      * @param salt The unique salt for deploying the token.
-     * @param localMinter TBD
-     * @param remoteMinter TBD
+     * @param remoteMinter The calldata representing the minter's address on the remote chain. Must be empty if no minter is specified.
+     * Reverts if the msg.sender does not have mint permission for the token.
      * @param destinationChain The name of the destination chain.
      * @param gasValue The amount of gas to send for the deployment.
      * @return tokenId The tokenId corresponding to the deployed InterchainToken.
      */
     function deployRemoteInterchainToken(
         bytes32 salt,
-        address localMinter,
-        address remoteMinter,
+        bytes memory remoteMinter,
         string memory destinationChain,
         uint256 gasValue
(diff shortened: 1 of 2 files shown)


contracts/InterchainTokenFactory.sol
```diff
@@ -181,14 +182,19 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
         tokenSymbol = token.symbol();
         tokenDecimals = token.decimals();
 
-        if (minter != address(0)) {
-            if (!token.isMinter(minter)) revert NotMinter(minter);
-            if (minter == address(interchainTokenService)) revert InvalidMinter(minter);
+        if (localMinter != address(0)) {
```
Foivos, line 185: What is the point of localMinter? You do some checks with it and do nothing with it, so there's not reason to not pass `address(0)` to the function.

ahramy: had discussion, updated PR

contracts/InterchainTokenFactory.sol
```diff
@@ -207,13 +204,14 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
...
-        address minter,
+        address /*minter*/,
         string memory destinationChain,
         uint256 gasValue
     ) external payable returns (bytes32 tokenId) {
         if (bytes(originalChainName).length != 0) revert NotSupported();
+        bytes memory minter_ = new bytes(0);
 
-        tokenId = deployRemoteInterchainToken(salt, minter, destinationChain, gasValue);
+        tokenId = deployRemoteInterchainToken(salt, minter_, destinationChain, gasValue);
```
Foivos, line 214: this is incosistent with the other function. I think we should not have the same function name twice in the same contract anyway, and this function does nothing for now, since the additional parameter has to be empty. If you wanna add the parameter now and support it later then maybe remove the other function completely. If you need both you should have an internal function that executes the logic and have both external functions call it to avoid the extra external call, which I think is what happens here (test this i might be wrong).

PR #298, opened 2024-11-07
fix(its): remove unnecessary use of payable in functions

ahramy:
[AXE-4706](https://axelarnetwork.atlassian.net/browse/AXE-4706)
- I tested all functions with use of payable. To verify, I removed payable and ran unit tests one by one. 

### InterchainTokenFactory.sol
[x] deployInterchainToken -> MulticallFailed, unit test failed
[x] deployRemoteInterchainToken -> MulticallFailed, unit test failed
[x] registerCanonicalInterchainToken -> MulticallFailed, TakeTokenFailed, unit test failed
[x] deployRemoteCanonicalInterchainToken -> MulticallFailed, unit test failed

### TokenHandler.sol
**[o] giveToken -> `payable` can be removed.**
[x] takeToken -> lockUnlockFee failed
**[o] transferTokenFrom -> `payable` can be removed.**
[x] postTokenManagerDeploy -> PostDeployFailed

### InterchainTokenService.sol
[x] deployTokenManager -> failed
[x] deployInterchainToken -> Cannot set option "value" on a non-payable function type.
[x] expressExecute 
[x] interchainTransfer -> MultiCallFailed, non-payable method cannot override value
**[o] callContractWithInterchainToken -> `payable` can be removed.** 
**=> Not sure we should remove it. Are we allowing ETH to be sent or not?**
=> Unit test should be added.

### TransmitInterchainToken.sol
[x] transmitInterchainTransfer -> Cannot set option "value" on a non-payable function type.

### InterchainTokenStandard.sol
[x] interchainTransfer -> "msg.value" and "callvalue()" can only be used in payable public functions.
[x] interchainTransferFrom -> "msg.value" and "callvalue()" can only be used in payable public functions.

### GatewayCaller.sol
[x] callContract -> GatewayCallFailed

### TokenManagerDeployer
[x] deployTokenManager -> TokenManagerDeploymentFailed

[AXE-4706]: https://axelarnetwork.atlassian.net/browse/AXE-4706?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/contracts/InterchainTokenFactory.sol
+++ b/contracts/InterchainTokenFactory.sol
@@ -104,7 +104,7 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
     /**
      * @notice Deploys a new interchain token with specified parameters.
      * @dev Creates a new token and optionally mints an initial amount to a specified minter.
-     * The function is `payable` because non-payable functions cannot be called in a multi-call context.
+     * This function is `payable` because non-payable functions cannot be called in a multicall that calls other `payable` functions.
      * @param salt The unique salt for deploying the token.
      * @param name The name of the token.
      * @param symbol The symbol of the token.
@@ -251,7 +251,7 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
 
     /**
      * @notice Registers a canonical token as an interchain token and deploys its token manager.
-     * The function is `payable` because non-payable functions cannot be called in a multi-call context.
+     * This function is `payable` because non-payable functions cannot be called in a multicall that calls other `payable` functions.
      * @param tokenAddress The address of the canonical token.
      * @return tokenId The tokenId corresponding to the registered canonical token.
      */
--- a/contracts/InterchainTokenFactory.sol
+++ b/contracts/InterchainTokenFactory.sol
@@ -104,7 +104,7 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
     /**
      * @notice Deploys a new interchain token with specified parameters.
      * @dev Creates a new token and optionally mints an initial amount to a specified minter.
-     * The function is marked as `payable` to enable multi-call execution.
+     * The function is `payable` because non-payable functions cannot be called in a multi-call context.
      * @param salt The unique salt for deploying the token.
      * @param name The name of the token.
      * @param symbol The symbol of the token.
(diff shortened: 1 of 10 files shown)


contracts/InterchainTokenService.sol
```diff
@@ -513,7 +513,7 @@ contract InterchainTokenService is
         uint256 amount,
         bytes memory data,
         uint256 gasValue
-    ) external payable whenNotPaused {
+    ) external whenNotPaused {
```
Foivos, line 516: this allows for this function to be part of multicalls

Foivos (commented):
You cannot call non payable functions in multical, which is why the payable is needed on all these functions.

test/InterchainTokenService.js
```diff
@@ -1687,7 +1687,9 @@ describe('Interchain Token Service', () => {
 
                 await expect(
                     reportGas(
-                        service.callContractWithInterchainToken(tokenId, destinationChain, destAddress, amount, data, 0),
+                        service.callContractWithInterchainToken(tokenId, destinationChain, destAddress, amount, data, 0, {
+                            value: gasValue,
```
ahramy, line 1691: Forcing `callContractWithInterchainToken` function as `payable` in unit test.

contracts/TokenHandler.sol
```diff
@@ -32,7 +32,7 @@ contract TokenHandler is ITokenHandler, ITokenManagerType, ReentrancyGuard, Crea
      * @return address the address of the token.
      */
     // slither-disable-next-line locked-ether
-    function giveToken(bytes32 tokenId, address to, uint256 amount) external payable returns (uint256, address) {
+    function giveToken(bytes32 tokenId, address to, uint256 amount) external returns (uint256, address) {
```
milapsheth, line 35: can you make sure out tests cover a call to `interchainTransfer` with non-zero gasValue so this gets triggered?

milapsheth (approved):

contracts/InterchainTokenFactory.sol
```diff
@@ -104,7 +104,7 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
     /**
      * @notice Deploys a new interchain token with specified parameters.
      * @dev Creates a new token and optionally mints an initial amount to a specified minter.
-     * The function is marked as `payable` to enable multi-call execution.
+     * The function is `payable` because non-payable functions cannot be called in a multi-call context.
```
milapsheth, line 107:
```suggestion
     * This function is `payable` because non-payable functions cannot be called in a multicall that calls other `payable` functions.
```
Still not quite right. update the rest

PR #300, opened 2024-11-09
feat(its): add validation checks for evm its message fields

ahramy:
[AXE-6320](https://axelarnetwork.atlassian.net/browse/AXE-6320)

[AXE-6320]: https://axelarnetwork.atlassian.net/browse/AXE-6320?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/contracts/InterchainTokenService.sol
+++ b/contracts/InterchainTokenService.sol
@@ -493,8 +493,6 @@ contract InterchainTokenService is
         bytes calldata metadata,
         uint256 gasValue
     ) external payable whenNotPaused {
-        if (destinationAddress.length == 0) revert EmptyDestinationAddress();
-
         amount = _takeToken(tokenId, msg.sender, amount, false);
 
         (IGatewayCaller.MetadataVersion metadataVersion, bytes memory data) = _decodeMetadata(metadata);
@@ -518,7 +516,6 @@ contract InterchainTokenService is
         bytes memory data,
         uint256 gasValue
     ) external payable whenNotPaused {
-        if (destinationAddress.length == 0) revert EmptyDestinationAddress();
         if (data.length == 0) revert EmptyData();
 
         amount = _takeToken(tokenId, msg.sender, amount, false);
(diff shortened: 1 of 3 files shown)


contracts/InterchainTokenService.sol
```diff
@@ -551,6 +557,9 @@ contract InterchainTokenService is
         uint256 amount,
         bytes calldata metadata
     ) external payable whenNotPaused {
+        if (sourceAddress == address(0)) revert EmptySourceAddress();
```
milapsheth, line 560: not really possible, and this is also not equivalent to sourceAddress bytes being empty (this would be converted to 20 zero bytes). So we can skip this check

contracts/InterchainTokenService.sol
```diff
@@ -551,6 +557,9 @@ contract InterchainTokenService is
...
+        if (destinationAddress.length == 0) revert EmptyDestinationAddress();
```
milapsheth, line 561: Why not check inside `_transmitInterchainTransfer` instead? Some code paths aren't being covered

milapsheth (approved):

contracts/InterchainTokenService.sol
ahramy: make sense, yes, it will be converted to 20 zero bytes. removed the check

contracts/InterchainTokenService.sol
ahramy: It was because `transmitInterchainTransfer` checks for `EmptySourceAddress()` and `EmptyDestinationAddress()`, which are not required in `_transmitInterchainTransfer`. I thought this might optimize the gas cost since checking the source address would not be necessary. But after removing the `EmptySourceAddress()` check from `transmitInterchainTransfer`, we can move the checkin inside `_transmitInterchainTransfer`. updated.

milapsheth (approved):

PR #301, opened 2024-11-10
feat(its-factory)!: support arbitrary destination minter for remote deployment

milapsheth:
[AXE-6319]

[AXE-6319]: https://axelarnetwork.atlassian.net/browse/AXE-6319?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/contracts/InterchainTokenFactory.sol
+++ b/contracts/InterchainTokenFactory.sol
@@ -18,13 +18,33 @@ import { IInterchainToken } from './interfaces/IInterchainToken.sol';
 contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, Multicall, Upgradable {
     using AddressBytes for address;
 
-    IInterchainTokenService public immutable interchainTokenService;
-    bytes32 public immutable chainNameHash;
+    /// @dev This slot contains the storage for this contract in an upgrade-compatible manner
+    /// keccak256('InterchainTokenFactory.Slot') - 1;
+    bytes32 internal constant INTERCHAIN_TOKEN_FACTORY_SLOT =
+        0xd4f5c43117c663161acfe6af3208a49856d85e586baf0f60749de2055e001465;
 
     bytes32 private constant CONTRACT_ID = keccak256('interchain-token-factory');
     bytes32 internal constant PREFIX_CANONICAL_TOKEN_SALT = keccak256('canonical-token-salt');
     bytes32 internal constant PREFIX_INTERCHAIN_TOKEN_SALT = keccak256('interchain-token-salt');
+    bytes32 internal constant PREFIX_DEPLOY_APPROVAL = keccak256('deploy-approval');
     address private constant TOKEN_FACTORY_DEPLOYER = address(0);
+
+    IInterchainTokenService public immutable interchainTokenService;
+    bytes32 public immutable chainNameHash;
+
+    struct DeployApproval {
+        address minter;
+        address deployer;
+        bytes32 salt;
+        string destinationChain;
+        bytes destinationMinter;
+    }
+
+    /// @dev Storage for this contract
+    /// @param deployApprovals Mapping of deployment approvals
+    struct InterchainTokenFactoryStorage {
+        mapping(bytes32 => bool) deployApprovals;
+    }
 
     /**
      * @notice Constructs the InterchainTokenFactory contract.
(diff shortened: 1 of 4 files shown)


contracts/InterchainTokenFactory.sol
```diff
@@ -152,19 +172,73 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
         }
     }
 
+    /**
+     * @notice Allow the minter to approve the deployer for a remote interchain token deployment that uses a custom destinationMinter address.
+     * This ensures that a token deployer can't choose the destinationMinter itself, and requires the approval of the minter to reduce trust assumptions on the deployer.
+     */
+    function approveDeployRemoteInterchainToken(address deployer, bytes32 salt, string calldata destinationChain, bytes calldata destinationMinter) external {
```
ahramy, line 179: Anyone can call the `approveDeployRemoteInterchainToken` function. Would it be safe to derive the `tokenId` from `salt` and check `if (!token.isMinter(msg.sender)) revert NotMinter(msg.sender);` to make the approval `true? This might be overkill since the `if (!token.isMinter(minter)) revert NotMinter(minter);` check is already present in the `deployRemoteInterchainToken` function, but I think it’s good to ensure there are no security concerns.

contracts/InterchainTokenFactory.sol
```diff
@@ -183,9 +257,27 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
...
+                });
+                _useDeployApproval(approval);
+                minter_ = destinationMinter;
+            } else {
+                minter_ = minter.toBytes();
+            }
+
+            // Sanity check to prevent accidental use of the current ITS address as the destination minter
+            bytes memory itsAddress = address(interchainTokenService).toBytes();
+            if (keccak256(minter_) == keccak256(itsAddress)) revert InvalidMinter(minter);
```
ahramy, line 277: I think the one-liner is self-explanatory, but two lines are also ok with me.
```
// Sanity check to prevent accidental use of the current ITS address as the destination minter
if (keccak256(minter_) == keccak256(address(interchainTokenService).toBytes())) revert InvalidMinter(minter);
```

contracts/InterchainTokenFactory.sol
milapsheth: fair point, we can add the check, and since the gas cost isn't a concern. What we should ensure is to not remove the same check in deploy* because the minter could be transferred later, and the approval is stale. i'll add a comment for it

ahramy (approved):

contracts/InterchainTokenFactory.sol
```diff
@@ -153,19 +172,84 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
         }
     }
 
+    /**
+     * @notice Allow the minter to approve the deployer for a remote interchain token deployment that uses a custom destinationMinter address.
+     * This ensures that a token deployer can't choose the destinationMinter itself, and requires the approval of the minter to reduce trust assumptions on the deployer.
+     */
+    function approveDeployRemoteInterchainToken(
```
milapsheth, line 179: add a revoke method

contracts/InterchainTokenFactory.sol
```diff
@@ -153,19 +172,84 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
...
+    /**
+     * @notice Deploys a remote interchain token on a specified destination chain.
+     * @param salt The unique salt for deploying the token.
+     * @param minter The address to use as the minter of the deployed token on the destination chain. If the destination chain is not EVM,
+     * then use the more generic `deployRemoteInterchainToken` function below that allows setting an arbitrary destination minter that was approved by the current minter.
+     * @param destinationChain The name of the destination chain.
+     * @param gasValue The amount of gas to send for the deployment.
+     * @return tokenId The tokenId corresponding to the deployed InterchainToken.
+     */
+    function deployRemoteInterchainToken(
```
Foivos, line 227: Avoid adding the same function name with different params. I know the compiler allows you to, but I find it very confusing for readability.

contracts/InterchainTokenFactory.sol
```diff
@@ -184,9 +268,25 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
 
         if (minter != address(0)) {
             if (!token.isMinter(minter)) revert NotMinter(minter);
+            // Sanity check to prevent accidental use of the current ITS address as the destination minter
             if (minter == address(interchainTokenService)) revert InvalidMinter(minter);
 
-            minter_ = minter.toBytes();
+            if (destinationMinter.length > 0) {
```
Foivos, line 274: So if I understand this correctly the minter can approve someone else (the 'deployer') to deploy a remote token to a specific chain with a specific minter. If this is true, why don't they just deploy it themselves?

PR #303, opened 2024-11-16
test(its-factory): add upgrade test

milapsheth:
[AXE-6587]

[AXE-6587]: https://axelarnetwork.atlassian.net/browse/AXE-6587?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/test/InterchainTokenFactory.js
+++ b/test/InterchainTokenFactory.js
@@ -51,7 +51,7 @@ describe('InterchainTokenFactory', () => {
         });
     });
 
-    describe.only('Upgrade', async () => {
+    describe('Upgrade', async () => {
         it('Should revert on upgrade from non-owner account', async () => {
             await expectRevert(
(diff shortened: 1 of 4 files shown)


test/InterchainTokenFactory.js
```diff
@@ -50,6 +51,40 @@ describe('InterchainTokenFactory', () => {
         });
     });
 
+    describe('Upgrade', async () => {
```
ahramy, line 54: Should we add an upgrade test to `InterchainTokenService` since it is also 'Upgradable'?

ahramy (approved):

test/InterchainTokenFactory.js
milapsheth: it has a dedicated upgrade test in `InterchainTokenServiceUpgradeFlow.js`

milapsheth: Might move the factory test there in the future, not sure

PR #311, opened 2024-12-10
feat(its): add custom token support with auto scaling

milapsheth:
- [ ] Update tests
- [ ] Move custom token usage from ITS to ITS factory tests
- [ ] Remove `deployTokenManager` to reduce ITS bytecode after tests are updated to use `linkToken`

Two new message types are introduced:
- `RegisterTokenMetadata`: Sends a message to ITS hub to register token decimals by token address
- `LinkToken`: Allows linking custom tokens across chains. LinkToken needs to be submitted from the same origin chain to other remote chains for the same `tokenId`. Existing custom tokens can still be linked by calling `linkToken` from a chain that has a registered token for that `tokenId`.


contracts/InterchainTokenService.sol
```diff
@@ -301,31 +314,38 @@ contract InterchainTokenService is
...
         uint256 gasValue
-    ) external payable whenNotPaused returns (bytes32 tokenId) {
-        if (bytes(params).length == 0) revert EmptyParams();
+    ) external payable returns (bytes32 tokenId) {
+        (bytes memory linkParams, address destinationTokenAddress) = abi.decode(params, (bytes, address));
 
+        tokenId = linkToken(salt, destinationChain, destinationTokenAddress.toBytes(), tokenManagerType, false, linkParams, gasValue);
+    }
+
+    function linkToken(bytes32 salt, string calldata destinationChain, bytes memory destinationTokenAddress, TokenManagerType tokenManagerType, bool autoScaling, bytes memory linkParams, uint256 gasValue) public payable whenNotPaused returns (bytes32 tokenId) {
```
Foivos, line 323: If this is to be used for local deployments too then linkToken is confusing.

contracts/InterchainTokenService.sol
```diff
@@ -301,31 +314,38 @@ contract InterchainTokenService is
...
         address deployer = msg.sender;
 
-        if (deployer == interchainTokenFactory) {
+        if (msg.sender == interchainTokenFactory) {
             deployer = TOKEN_FACTORY_DEPLOYER;
-        } else if (bytes(destinationChain).length == 0 && trustedAddressHash(chainName()) == ITS_HUB_ROUTING_IDENTIFIER_HASH) {
-            // Restricted on ITS contracts deployed to Amplifier chains until ITS Hub adds support
-            revert NotSupported();
+        } else if (bytes(destinationChain).length == 0) {
+            // TODO: Only support linking new tokens via ITS factory, to include chain name in token id derivation
```
Foivos, line 335: What happens to already deployed custom tokens?

contracts/InterchainTokenService.sol
```diff
@@ -731,15 +751,15 @@ contract InterchainTokenService is
...
+        (, bytes32 tokenId, TokenManagerType tokenManagerType, , bytes memory destinationTokenAddress, , bytes memory linkParams) = abi.decode(
             payload,
-            (uint256, bytes32, TokenManagerType, bytes)
+            (uint256, bytes32, TokenManagerType, bytes, bytes, bool, bytes)
         );
 
         if (tokenManagerType == TokenManagerType.NATIVE_INTERCHAIN_TOKEN) revert CannotDeploy(tokenManagerType);
 
-        _deployTokenManager(tokenId, tokenManagerType, params);
+        _deployTokenManager(tokenId, tokenManagerType, destinationTokenAddress.toAddress(), linkParams);
```
Foivos, line 762: I think even if the only param left in EVM is the operator, we should have it be abi encoded when it is passed as params.

contracts/InterchainTokenFactory.sol
```diff
@@ -446,6 +447,32 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
         tokenId = deployRemoteCanonicalInterchainToken(originalTokenAddress, destinationChain, gasValue);
     }
 
+    // TODO: Should we reuse interchainTokenDeploySalt?
```
milapsheth, line 450:
```suggestion
```
removed since it can be problematic for trustlessly deployed tokens if the deployer key is compromised

PR #312, opened 2024-12-10
feat: addressed ackee's report

Foivos:


contracts/InterchainTokenFactory.sol
```diff
@@ -149,6 +149,8 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
             if (minter == address(interchainTokenService)) revert InvalidMinter(minter);
 
             minterBytes = minter.toBytes();
+        } else {
+            revert EmptyInterchainToken();
```
milapsheth, line 153:
```suggestion
            revert ZeroSupplyToken();
```
maybe?

milapsheth: also mention this behaviour in the docstring

contracts/InterchainTokenService.sol
```diff
@@ -395,6 +395,26 @@ contract InterchainTokenService is
         return _contractCallValue(payload);
     }
 
+    /**
+     * @notice Executes operations based on the payload and selector.
```
milapsheth, line 399:
```suggestion
     * @notice Executes the cross-chain ITS message.
```
unrelated to the issue but noticed an improvement

contracts/TokenHandler.sol
```diff
@@ -16,12 +16,11 @@ import { IERC20BurnableFrom } from './interfaces/IERC20BurnableFrom.sol';
```
milapsheth, line 16: fix above lint

test/InterchainToken.js
```diff
@@ -147,7 +147,7 @@ describe('InterchainToken', () => {
             const contractBytecodeHash = keccak256(contractBytecode);
 
             const expected = {
-                london: '0xa01cf28b0b6ce6dc3b466e995585d69486400d671fce0ea8d06beba583e6f3bb',
+                london: '0x482146829055f052063003e9cf0ffaf798a12fb58088c2667566a135b9568355',
```
milapsheth, line 150: Avoid changing contracts with fixed bytecode unless necessary. We want to avoid multiple versions of tokens to be active at the same time since we can't upgrade previous ones

contracts/interchain-token/InterchainToken.sol
```diff
@@ -16,9 +13,7 @@ import { Minter } from '../utils/Minter.sol';
  * @notice This contract implements an interchain token which extends InterchainToken functionality.
  * @dev This contract also inherits Minter and Implementation logic.
  */
-contract InterchainToken is InterchainTokenStandard, ERC20, ERC20Permit, Minter, IInterchainToken {
-    using AddressBytes for bytes;
-
+contract InterchainToken is InterchainTokenStandard, ERC20Permit, Minter, IInterchainToken {
```
milapsheth, line 16: we don't want to change token bytecode unless necessary

test/TokenManager.js
```diff
@@ -92,7 +92,7 @@ describe('Token Manager', () => {
             const proxyBytecodeHash = keccak256(proxyBytecode);
 
             const expected = {
-                london: '0x8080880884e00735cc1a34bdf5c1ea6c023db60a01cfa1e951ca41ecf5fd8836',
+                london: '0x3b336208cc75ca67bdd39bdeed72871ce795e6e9cd28e20f811599ea51973ebf',
```
milapsheth, line 95: same proxy code shouldn't be changed

test/InterchainTokenFactory.js
```diff
@@ -235,18 +235,13 @@ describe('InterchainTokenFactory', () => {
         it('Should register a token if the mint amount is zero and minter is the zero address', async () => {
```
milapsheth, line 235: update test name

test/InterchainToken.js
```diff
@@ -147,7 +147,7 @@ describe('InterchainToken', () => {
             const contractBytecodeHash = keccak256(contractBytecode);
 
             const expected = {
-                london: '0x482146829055f052063003e9cf0ffaf798a12fb58088c2667566a135b9568355',
+                london: '0xa01cf28b0b6ce6dc3b466e995585d69486400d671fce0ea8d06beba583e6f3bb',
```
milapsheth, line 150: prevent changing this bytecode since it's not necessary

Foivos: this is changed back to the original bytecode

contracts/InterchainTokenFactory.sol
```diff
@@ -404,10 +404,24 @@ contract InterchainTokenFactory is IInterchainTokenFactory, ITokenManagerType, M
         bytes32 deploySalt = canonicalInterchainTokenDeploySalt(tokenAddress);
         string memory currentChain = '';
         uint256 gasValue = 0;
+        _checkToken(tokenAddress);
 
         tokenId = interchainTokenService.deployTokenManager(deploySalt, currentChain, TokenManagerType.LOCK_UNLOCK, params, gasValue);
     }
 
+    function _checkToken(address tokenAddress) internal view {
```
milapsheth, line 412: create a helper `_getTokenMetadata` that returns name, symbol, decimals and has a single try catch that reverts with NotToken if this fails. This helper can then be used in other functions in this contract for consistency, since the token could be upgraded between deployment calls

contracts/TokenHandler.sol
```diff
@@ -16,12 +16,11 @@ import { IERC20BurnableFrom } from './interfaces/IERC20BurnableFrom.sol';
 
 /**
```
milapsheth, line 17: fix unused SafeTokenTransfer import lint warning

contracts/InterchainTokenFactory.sol
Foivos: The reason why that's not optimal is because putting `name` and `symbol` in memory sill use gas even if the values are not used after. Also try-catch are only available in external calls, so to do what you are suggesting I would need to make the call external, which is even more gas, and more gas for other uses as well. Given the above, I think this would be a bad idea, lmk if you wanna do it anyway.

milapsheth: Since it's a deployment method, a bit more gas is fine to make this more consistent whenever metadata is retrieved across multiple functions. You can keep the 3 try-catches since it can't be used on an internal method

Foivos: So are you suggesting we do it in an internal method with the try-catches then? It is not very clear to me.

milapsheth (approved):

PR #316, opened 2025-01-06
feat(its): allow its owner to manually migrate tokens

Foivos:
[AXE-7067]

[AXE-7067]: https://axelarnetwork.atlassian.net/browse/AXE-7067?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


contracts/InterchainTokenService.sol
```diff
@@ -623,6 +624,29 @@ contract InterchainTokenService is
         }
     }
 
+    /**
+     * @notice Allows the owner to migrate legacy tokens that cannot be migrated automatically.
+     * @param tokenId the tokenId of the registered token.
+     * @param data the data to pass to migrate the token.
+     */
+    function migrateLegacyToken(bytes32 tokenId, bytes calldata data) external onlyOwner {
```
milapsheth, line 632: So these were the legacy custom tokens:
https://interoplabs.slack.com/archives/C05F50E50MP/p1727995328548309?thread_ts=1727798397.640599&cid=C05F50E50MP

And all of these expose OZ's `grantRole` entrypoint, but it can only be called by the admin, and so wouldn't be callable by ITS (which would only have the minter role). So, I think we'll just need to coordinate with the partners to transfer the mint role and not need this method.

contracts/InterchainTokenService.sol
```diff
@@ -623,6 +624,29 @@ contract InterchainTokenService is
...
+        address tokenAddress = registeredTokenAddress(tokenId);
+        (bool success, bytes memory returnData) = tokenAddress.call(data);
+        if (!success) {
+            revert TokenMigrateFailed(returnData);
+        }
+    }
+
+    /**
+     * @notice Allows the owner to migrate legacy tokens that cannot be migrated automatically.
```
milapsheth, line 641:
```suggestion
     * @notice Allows the owner to migrate minter of native interchain tokens from ITS to the corresponding token manager.
```

contracts/InterchainTokenService.sol
```diff
@@ -623,6 +624,29 @@ contract InterchainTokenService is
...
+     * @param tokenId the tokenId of the registered token.
+     */
+    function migrateInterchainToken(bytes32 tokenId) external onlyOwner {
+        ITokenManager tokenManager_ = deployedTokenManager(tokenId);
+        address tokenAddress = tokenManager_.tokenAddress();
+        IMinter(tokenAddress).transferMintership(address(tokenManager_));
```
milapsheth, line 647: for unit test, also test failure when migrating an already migrated token

contracts/interfaces/IInterchainTokenService.sol
```diff
@@ -275,4 +276,17 @@ interface IInterchainTokenService is
...
+
+    /**
+     * @notice Allows the owner to migrate legacy tokens that cannot be migrated automatically.
+     * @param tokenId the tokenId of the registered token.
+     * @param data the data to pass to migrate the token.
+     */
+    function migrateLegacyToken(bytes32 tokenId, bytes calldata data) external;
+
+    /**
+     * @notice Allows the owner to migrate legacy tokens that cannot be migrated automatically.
```
milapsheth, line 288: same

contracts/InterchainTokenService.sol
Foivos: removed

contracts/InterchainTokenService.sol
Foivos: added all tests, please check

contracts/interfaces/IInterchainTokenService.sol
Foivos: remove this method

milapsheth (approved):

PR #332, opened 2025-03-25
feat: make all calls go through the hub

Foivos:
[AXE-8091]

[AXE-8091]: https://axelarnetwork.atlassian.net/browse/AXE-8091?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


contracts/InterchainTokenService.sol
```diff
@@ -785,68 +785,30 @@ contract InterchainTokenService is
         IGatewayCaller.MetadataVersion metadataVersion,
```
ahramy, line 785: nit: in the comment, trustedAddress -> isTrustedChain ?

contracts/interfaces/IInterchainChainTracker.sol
```diff
@@ -0,0 +1,15 @@
...
+interface IInterchainChainTracker {
+    event TrustedChainSet(string chainName);
+    event TrustedChainRemoved(string chainName);
+    event ItsHubAddressSet(string hubAddress);
+
+    function itsHubAddress() external view returns (string memory hubAddress);
+
+    function itsHubAddressHash() external view returns (bytes32 hubAddressHash);
+
+    function isTrustedChain(string memory chain) external view returns (bool trusted);
```
ahramy, line 14: same as above

contracts/interfaces/IInterchainChainTracker.sol
```diff
@@ -0,0 +1,15 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.0;
+
+interface IInterchainChainTracker {
+    event TrustedChainSet(string chainName);
```
ahramy, line 6: `setTrustedChain` and `removeTrustedChain` take variable name `chain`. Here it takes `chainName`. Maybe update `chainName` to `chain` or the other way around for consistency in the code.

contracts/InterchainTokenService.sol
```diff
@@ -171,7 +166,8 @@ contract InterchainTokenService is
      * @param sourceAddress The source address that the call came from.
      */
     modifier onlyRemoteService(string calldata sourceChain, string calldata sourceAddress) {
```
milapsheth, line 168:
```suggestion
    modifier onlyITSHub(string calldata sourceChain, string calldata sourceAddress) {
```
also update the docstring

contracts/InterchainTokenService.sol
```diff
@@ -171,7 +166,8 @@ contract InterchainTokenService is
...
-        if (!isTrustedAddress(sourceChain, sourceAddress)) revert NotRemoteService();
+        if (keccak256(bytes(sourceChain)) != ITS_HUB_CHAIN_NAME_HASH || keccak256(bytes(sourceAddress)) != itsHubAddressHash())
+            revert NotRemoteService();
```
milapsheth, line 170:
```suggestion
            revert NotITSHub();
```

contracts/InterchainTokenService.sol
```diff
@@ -456,31 +446,37 @@ contract InterchainTokenService is
      * @notice Express executes operations based on the payload and selector.
      * @dev This function is `payable` because non-payable functions cannot be called in a multicall that calls other `payable` functions.
      * @param commandId The unique message id.
-     * @param sourceChain The chain where the transaction originates from.
```
milapsheth, line 459: why remove this?

contracts/InterchainTokenService.sol
```diff
@@ -663,21 +658,26 @@ contract InterchainTokenService is
...
         _addOperator(operator);
         _setChainName(chainName_);
+        _setItsHubAddress(itsHubAddress_);
 
+        uint256 length = trustedChainNames.length;
         for (uint256 i; i < length; ++i) {
-            _setTrustedAddress(trustedChainNames[i], trustedAddresses[i]);
+            string memory trustedChainName = trustedChainNames[i];
+            _setTrustedChain(trustedChainName);
+            // Remove previously set trusted addresses.
```
milapsheth, line 677:
```suggestion

            // Migration: Remove previously set trusted addresses.
```
we should add a dedicated `_migrate` method in the future that's only run when not doing the initial deployment

contracts/InterchainTokenService.sol
```diff
@@ -785,68 +785,30 @@ contract InterchainTokenService is
...
-    function _callContract(
-        string memory destinationChain,
-        string memory destinationAddress,
-        bytes memory payload,
-        IGatewayCaller.MetadataVersion metadataVersion,
-        uint256 gasValue
-    ) internal {
-        // Check whether no trusted address was set for the destination chain
-        if (bytes(destinationAddress).length == 0) revert UntrustedChain();
+    function _callContract(bytes memory payload, IGatewayCaller.MetadataVersion metadataVersion, uint256 gasValue) internal {
```
milapsheth, line 802:
```suggestion
    function _sendToHub(bytes memory payload, IGatewayCaller.MetadataVersion metadataVersion, uint256 gasValue) internal {
```
rename for clarity?

contracts/InterchainTokenService.sol
```diff
@@ -785,68 +785,30 @@ contract InterchainTokenService is
...
         (bool success, bytes memory returnData) = gatewayCaller.delegatecall(
-            abi.encodeWithSelector(
-                IGatewayCaller.callContract.selector,
-                destinationChain,
-                destinationAddress,
-                payload,
-                metadataVersion,
-                gasValue
-            )
+            abi.encodeWithSelector(IGatewayCaller.callContract.selector, ITS_HUB_CHAIN_NAME, hubAddress, payload, metadataVersion, gasValue)
```
milapsheth, line 806: Maybe in a follow up PR, we could remove the gateway caller util? We don't have gateway token support anymore, so the extra gas cost and complexity seems unnecessary.
https://axelarnetwork.atlassian.net/browse/AXE-8398

contracts/InterchainTokenService.sol
```diff
@@ -893,20 +855,19 @@ contract InterchainTokenService is
         string memory originalSourceChain = sourceChain;
```
milapsheth, line 855:
```suggestion
        string memory originalSourceChain;
```
doesn't need to be initialized to sourceChain anymore

contracts/InterchainTokenService.sol
```diff
@@ -893,20 +855,19 @@ contract InterchainTokenService is
...
 
         // Unwrap ITS message if coming from ITS hub
-        if (messageType == MESSAGE_TYPE_RECEIVE_FROM_HUB) {
-            if (keccak256(abi.encodePacked(sourceChain)) != ITS_HUB_CHAIN_NAME_HASH) revert UntrustedChain();
+        if (messageType != MESSAGE_TYPE_RECEIVE_FROM_HUB) {
+            revert InvalidMessageType(messageType);
+        }
 
-            (, originalSourceChain, payload) = abi.decode(payload, (uint256, string, bytes));
+        if (keccak256(bytes(sourceChain)) != ITS_HUB_CHAIN_NAME_HASH) revert UntrustedChain();
```
milapsheth, line 862: won't the `onlyRemoteService` modifier check this already?

contracts/InterchainTokenService.sol
```diff
@@ -785,68 +785,30 @@ contract InterchainTokenService is
...
-            destinationChain = ITS_HUB_CHAIN_NAME;
-            destinationAddress = trustedAddress(ITS_HUB_CHAIN_NAME);
-        }
-
-        return (destinationChain, destinationAddress, payload);
-    }
-
     function _execute(
         bytes32 commandId,
         string calldata sourceChain,
```
milapsheth, line 814: `_getExecuteParams` -> `_decodeHubMessage` maybe?

contracts/InterchainTokenService.sol
```diff
@@ -1120,8 +1081,15 @@ contract InterchainTokenService is
      * @return address The token address.
      * @return uint256 The value the call is worth.
      */
-    function _contractCallValue(bytes calldata payload) internal view returns (address, uint256) {
-        (uint256 messageType, bytes32 tokenId, , , uint256 amount) = abi.decode(payload, (uint256, bytes32, bytes, bytes, uint256));
+    function _contractCallValue(bytes memory payload) internal view returns (address, uint256) {
+        uint256 messageType;
+        (messageType, , payload) = abi.decode(payload, (uint256, string, bytes));
```
milapsheth, line 1086: reuse `_getExecuteParams` to avoid duplication. Gas cost isn't an issue since this only gets queried by the express relayer

contracts/InterchainTokenService.sol
```diff
@@ -456,31 +446,37 @@ contract InterchainTokenService is
...
-        uint256 messageType = abi.decode(payload, (uint256));
+        bytes32 payloadHash = keccak256(payload);
+        uint256 messageType;
+        string memory originalSourceChain;
+        (messageType, originalSourceChain, payload) = abi.decode(payload, (uint256, string, bytes));
+        if (messageType != MESSAGE_TYPE_RECEIVE_FROM_HUB) {
+            revert InvalidExpressMessageType(messageType);
+        }
+
+        messageType = abi.decode(payload, (uint256));
```
milapsheth, line 466: use `_getMessageType`?

contracts/InterchainTokenService.sol
```diff
@@ -456,31 +446,37 @@ contract InterchainTokenService is
...
-        bytes calldata payload
+        bytes memory payload
     ) public payable whenNotPaused {
-        uint256 messageType = abi.decode(payload, (uint256));
+        bytes32 payloadHash = keccak256(payload);
+        uint256 messageType;
+        string memory originalSourceChain;
+        (messageType, originalSourceChain, payload) = abi.decode(payload, (uint256, string, bytes));
+        if (messageType != MESSAGE_TYPE_RECEIVE_FROM_HUB) {
+            revert InvalidExpressMessageType(messageType);
```
milapsheth, line 463: reuse `_getExecuteParams`

test/AddressDerivation.js
```diff
@@ -12,7 +12,13 @@ const { approveContractCall } = require('../scripts/utils');
 const { getRandomBytes32, getSaltFromKey, isHardhat, getContractJSON } = require('./utils');
 const { create3DeployContract } = require('@axelar-network/axelar-gmp-sdk-solidity');
 const Token = getContractJSON('TestInterchainTokenStandard');
-const { NATIVE_INTERCHAIN_TOKEN, MESSAGE_TYPE_DEPLOY_INTERCHAIN_TOKEN } = require('./constants');
+const {
+    NATIVE_INTERCHAIN_TOKEN,
+    MESSAGE_TYPE_DEPLOY_INTERCHAIN_TOKEN,
+    MESSAGE_TYPE_RECEIVE_FROM_HUB,
+    ITS_HUB_ADDRESS,
+    ITS_HUB_CHAIN_NAME,
```
milapsheth, line 20:
```suggestion
    ITS_HUB_CHAIN,
```

test/AddressDerivation.js
```diff
@@ -127,10 +141,14 @@ if (isHardhat) {
                     ['uint256', 'bytes32', 'string', 'string', 'uint8', 'bytes'],
                     [MESSAGE_TYPE_DEPLOY_INTERCHAIN_TOKEN, tokenId, tokenName, tokenSymbol, tokenDecimals, minter],
                 );
+                const wrappedPayload = defaultAbiCoder.encode(
+                    ['uint256', 'string', 'bytes'],
+                    [MESSAGE_TYPE_RECEIVE_FROM_HUB, sourceChain, payload],
+                );
```
milapsheth, line 147: create a helper for encoding ITS hub send and receive message for easier refactoring in the future
```suggestion
                const hubMessage = receiveFromHubMessage(sourceChain, payload);
```

Similarly, helpers from other message types would be convenient too

test/InterchainTokenFactory.js
```diff
@@ -140,6 +139,10 @@ describe('InterchainTokenFactory', () => {
                 ['uint256', 'bytes32', 'string', 'string', 'uint8', 'bytes'],
                 [MESSAGE_TYPE_DEPLOY_INTERCHAIN_TOKEN, tokenId, name, symbol, decimals, '0x'],
             );
+            const wrappedPayload = defaultAbiCoder.encode(
+                ['uint256', 'string', 'bytes'],
+                [MESSAGE_TYPE_SEND_TO_HUB, destinationChain, payload],
+            );
```
milapsheth, line 145:
```suggestion
            const hubMessage = sendHubMessage(destinationChain, payload);
```
same

contracts/utils/InterchainChainTracker.sol
```diff
@@ -0,0 +1,70 @@
...
+    bytes32 private constant ITS_HUB_ADDRESS_HASH_SLOT = 0x0c1d88a695a4dde782068c5771da4ade87fc66ae5dbbcc7d911d5557bc586c21;
+
+    bytes32 internal constant PREFIX_CHAIN_MAPPING = keccak256('interchain-chain-tracker-chain-mapping');
+
+    /**
+     * @dev Gets the key for the trusted address at a remote chain
+     * @param chain Chain name of the remote chain
+     * @return slot the slot to store the trusted address in
+     */
+    function _getTrustedChainSlot(string memory chain) internal pure returns (bytes32 slot) {
```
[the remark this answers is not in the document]
Foivos: done

contracts/InterchainTokenService.sol
Foivos: I feel this would be inelegant here: what would this `_migrate` function take as input? If we want to be generic it would be the `bytes params` which would be re-decoded and the chains would be iterated over again, to do a simple check. Alternatively, migrate could take in a chain name and only do 
```
if (trustedAddressHash(trustedChainName) != 0) {
    _removeTrustedAddress(trustedChainName);
}
```
which is fine but not meaningful for the future. What do you think?

contracts/InterchainTokenService.sol
Foivos: The gateway caller serves mainly to reduce contract size, but maybe now we can fit it into ITS, worth trying but unlikely.

contracts/InterchainTokenService.sol
Foivos: nice catch.

contracts/InterchainTokenService.sol
Foivos: nice catch.

contracts/InterchainTokenService.sol
Foivos: done by mistake, it is back.

contracts/InterchainTokenService.sol
milapsheth: yeah, we'll need to also change how the `Upgradable` works. Currently `setup` gets called both in the proxy constructor and during `upgrade`. However, `upgrade` might require dedicated migration code that shouldn't be run in the constructor. It's not been a major issue yet since we haven't needed to do many migrations. It's still not a major issue here so it's fine as-is. In the future, we could add a separate `migrate` method to `Upgradable` that gets called during `upgrade` whereas `setup` only gets called from the constructor, and the logic can differ then (e.g. `setup` doesn't need to remove trusted addresses if one exists)

contracts/InterchainTokenService.sol
```diff
@@ -127,10 +129,11 @@ contract InterchainTokenService is
         address gasService_,
         address interchainTokenFactory_,
         string memory chainName_,
+        string memory itsHubAddress_,
```
milapsheth, line 132: update docstring above

contracts/InterchainTokenService.sol
```diff
@@ -774,7 +769,7 @@ contract InterchainTokenService is
     /**
      * @notice Route the ITS message to the destination chain with the given payload
      * @dev This method also determines whether the ITS call should be routed via the ITS Hub.
```
milapsheth, line 771:
```suggestion
     * @notice Route the ITS message to the destination chain with the given payload via the ITS Hub.
```

contracts/InterchainTokenService.sol
```diff
@@ -774,7 +769,7 @@ contract InterchainTokenService is
...
-     * If the `trustedAddress(destinationChain) == 'hub'`, then the call is wrapped and routed to the ITS Hub destination.
+     * Only routes calls that return true for isTrustedChain(destinationChain)
      * @param destinationChain The target chain where the contract will be called.
```
milapsheth, line 773:
```suggestion
     * @param destinationChain The target chain where the contract will be called. The destinationChain must be set as trusted.
```

contracts/InterchainTokenService.sol
```diff
@@ -844,23 +839,18 @@ contract InterchainTokenService is
...
-        string calldata sourceChain,
-        bytes memory payload
-    ) internal view returns (uint256, string memory, bytes memory) {
+    function _decodeHubMessage(bytes memory payload) internal view returns (uint256, string memory, bytes memory) {
         // Read the first 32 bytes of the payload to determine the message type
         uint256 messageType = _getMessageType(payload);
 
-        // True source chain, this is overridden if the ITS call is coming via the ITS hub
-        string memory originalSourceChain = sourceChain;
+        // True source chain, this is found in the HUB payload.
```
milapsheth, line 846:
```suggestion
        // True source chain, this is found in the ITS Hub message.
```

contracts/InterchainTokenService.sol
```diff
@@ -844,23 +839,18 @@ contract InterchainTokenService is
...
+        string memory originalSourceChain;
 
         // Unwrap ITS message if coming from ITS hub
         if (messageType != MESSAGE_TYPE_RECEIVE_FROM_HUB) {
             revert InvalidMessageType(messageType);
         }
 
-        if (keccak256(bytes(sourceChain)) != ITS_HUB_CHAIN_NAME_HASH) revert UntrustedChain();
-
         (, originalSourceChain, payload) = abi.decode(payload, (uint256, string, bytes));
```
milapsheth, line 854:
```suggestion
        string memory originalSourceChain;
        (, originalSourceChain, payload) = abi.decode(payload, (uint256, string, bytes));
```
move it closer to where it's initialized

contracts/interfaces/IChainTracker.sol
```diff
@@ -4,14 +4,22 @@ pragma solidity ^0.8.0;
...
 
-import { IInterchainChainTracker } from './IInterchainChainTracker.sol';
-
 /**
  * @title IAddressTracker Interface
  * @notice This interface allows setting and removing a trusted address for a specific chain.
  * @dev Extends the IInterchainAddressTracker interface.
  */
-interface IChainTracker is IInterchainAddressTracker, IInterchainChainTracker {
+interface IChainTracker is IInterchainAddressTracker {
```
milapsheth, line 12:
```suggestion
interface IChainTracker {
```
keep address tracker separate?

contracts/interfaces/IItsHubAddressTracker.sol
```diff
@@ -0,0 +1,26 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.0;
+
+/**
+ * @title IAddressTracker Interface
+ * @notice This interface allows setting and removing a trusted address for a specific chain.
+ * @dev Extends the IInterchainAddressTracker interface.
```
milapsheth, line 8: update interface names/comments

contracts/interfaces/IItsHubAddressTracker.sol
```diff
@@ -0,0 +1,26 @@
...
+ */
+interface IItsHubAddressTracker {
+    error InvalidHubAddress();
+
+    /**
+     * @notice Getter for the ITS HUB address.
+     * @dev Needs to be overwitten.
+     * @return hubAddress The ITS HUB address.
```
milapsheth, line 16:
```suggestion
     * @notice Getter for the ITS Hub address.
     * @dev Needs to be overwritten.
     * @return hubAddress The ITS Hub address.
```

contracts/interfaces/IItsHubAddressTracker.sol
```diff
@@ -0,0 +1,26 @@
...
+     */
+    function itsHubAddress() external view returns (string memory hubAddress);
+
+    /**
+     * @notice Getter for the ITS HUB address hash.
+     * @dev Needs to be overwitten.
+     * @return hubAddressHash The ITS HUB address hash.
```
milapsheth, line 23:
```suggestion
     * @notice Getter for the ITS Hub address hash.
     * @dev Needs to be overwritten.
     * @return hubAddressHash The ITS Hub address hash.
```

contracts/utils/ChainTracker.sol
```diff
@@ -0,0 +1,38 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.0;
+
+import { IChainTracker } from '../interfaces/IChainTracker.sol';
+
+abstract contract ChainTracker is IChainTracker {
+    // uint256(keccak256('chain_tracker_slot')) - 1
```
milapsheth, line 8: use the same derivation as other contracts?
https://github.com/axelarnetwork/axelar-gmp-sdk-solidity/blob/main/contracts/gateway/BaseAmplifierGateway.sol#L11

contracts/utils/ChainTracker.sol
```diff
@@ -0,0 +1,38 @@
...
+    bytes32 internal constant CHAIN_TRACKER_SLOT = 0x6c51f1695661c243e1b0c3337f0f4c4f45116119470fb2cb0eb05fd857f16ac2;
```
milapsheth, line 9: create a `TestChainTracker` contract to test the hash derivation like we do for other contracts?

contracts/utils/ChainTracker.sol
```diff
@@ -0,0 +1,38 @@
...
+
+    struct ChainTrackerStorage {
+        mapping(string => bool) trustedChains;
+    }
+
+    function _setTrustedChain(string memory chainName) internal {
+        _chainTrackerStorage().trustedChains[chainName] = true;
+        emit TrustedChainSet(chainName);
```
milapsheth, line 17:
```suggestion
        _chainTrackerStorage().trustedChains[chainName] = true;

        emit TrustedChainSet(chainName);
```

contracts/utils/ChainTracker.sol
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    function _removeTrustedChain(string memory chainName) internal {
+        _chainTrackerStorage().trustedChains[chainName] = false;
+        emit TrustedChainRemoved(chainName);
```
milapsheth, line 22:
```suggestion

        emit TrustedChainRemoved(chainName);
```

contracts/utils/ItsHubAddressTracker.sol
```diff
@@ -0,0 +1,43 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.0;
+
+import { IItsHubAddressTracker } from '../interfaces/IItsHubAddressTracker.sol';
+
+abstract contract ItsHubAddressTracker is IItsHubAddressTracker {
```
milapsheth, line 7: add a docstring and explaining the motivation behind this, what the exact gas tradeoff is

contracts/utils/ItsHubAddressTracker.sol
```diff
@@ -0,0 +1,43 @@
...
+    bytes32 private immutable itsHubAddressMiddle;
+    uint8 private immutable itsHubAddressSuffix;
+
+    constructor(string memory hubAddress) {
+        if (bytes(hubAddress).length != 65) revert InvalidHubAddress();
+        bytes32 itsHubAddressPrefix_;
+        bytes32 itsHubAddressMiddle_;
+        uint8 itsHubAddressSuffix_;
+        assembly {
+            itsHubAddressPrefix_ := mload(add(hubAddress, 32))
```
milapsheth, line 22: create meaningful constants instead of hardcoding 32, 64, 65?

contracts/utils/ItsHubAddressTracker.sol
```diff
@@ -0,0 +1,43 @@
...
+    bytes32 private immutable itsHubAddressPrefix;
+    bytes32 private immutable itsHubAddressMiddle;
+    uint8 private immutable itsHubAddressSuffix;
+
+    constructor(string memory hubAddress) {
+        if (bytes(hubAddress).length != 65) revert InvalidHubAddress();
+        bytes32 itsHubAddressPrefix_;
+        bytes32 itsHubAddressMiddle_;
+        uint8 itsHubAddressSuffix_;
+        assembly {
```
milapsheth, line 21:
```suggestion
        if (bytes(hubAddress).length != 65) revert InvalidHubAddress();

        bytes32 itsHubAddressPrefix_;
        bytes32 itsHubAddressMiddle_;
        uint8 itsHubAddressSuffix_;

        assembly {
```
add newlines between different groups

contracts/utils/ItsHubAddressTracker.sol
```diff
@@ -0,0 +1,43 @@
...
+            itsHubAddressPrefix_ := mload(add(hubAddress, 32))
+            itsHubAddressMiddle_ := mload(add(hubAddress, 64))
+            itsHubAddressSuffix_ := mload(add(hubAddress, 65))
+        }
+        itsHubAddressPrefix = itsHubAddressPrefix_;
+        itsHubAddressMiddle = itsHubAddressMiddle_;
+        itsHubAddressSuffix = itsHubAddressSuffix_;
+        itsHubAddressHash = keccak256(bytes(hubAddress));
```
milapsheth, line 29:
```suggestion
        }

        itsHubAddressPrefix = itsHubAddressPrefix_;
        itsHubAddressMiddle = itsHubAddressMiddle_;
        itsHubAddressSuffix = itsHubAddressSuffix_;

        itsHubAddressHash = keccak256(bytes(hubAddress));
```

contracts/utils/ItsHubAddressTracker.sol
milapsheth, line 21: should we label assembly blocks as [memory-safe](https://docs.soliditylang.org/en/latest/assembly.html#memory-safety)?

contracts/utils/ItsHubAddressTracker.sol
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    function itsHubAddress() public view returns (string memory hubAddress) {
+        bytes32 itsHubAddressPrefix_ = itsHubAddressPrefix;
+        bytes32 itsHubAddressMiddle_ = itsHubAddressMiddle;
+        uint8 itsHubAddressSuffix_ = itsHubAddressSuffix;
+        hubAddress = new string(65);
+        assembly {
```
milapsheth, line 37:
```suggestion

        assembly {
```

contracts/utils/ItsHubAddressTracker.sol
```diff
@@ -0,0 +1,43 @@
...
+            mstore(add(hubAddress, 32), itsHubAddressPrefix_)
+            mstore(add(hubAddress, 65), itsHubAddressSuffix_)
+            mstore(add(hubAddress, 64), itsHubAddressMiddle_)
```
milapsheth, line 40: add a comment explaining why this is the order of operations since the third mstore will overwrite all but the last char. This will be helpful when someone looks at this in the future

contracts/InterchainTokenFactory.sol
```diff
@@ -192,7 +192,7 @@ contract InterchainTokenFactory is IInterchainTokenFactory, Multicall, Upgradabl
```
milapsheth, line 192: gh actions are failing

contracts/InterchainTokenService.sol
Foivos: ok, will leave as is but keep in mind. When making this I was surprised that setup was used for both as well.

contracts/InterchainTokenService.sol
```diff
@@ -658,23 +658,18 @@ contract InterchainTokenService is
...
-        _setItsHubAddress(itsHubAddress_);
 
         uint256 length = trustedChainNames.length;
         for (uint256 i; i < length; ++i) {
             string memory trustedChainName = trustedChainNames[i];
             _setTrustedChain(trustedChainName);
-            // Remove previously set trusted addresses.
+            // Migration: Remove previously set trusted addresses.
             if (trustedAddressHash(trustedChainName) != 0) {
                 _removeTrustedAddress(trustedChainName);
```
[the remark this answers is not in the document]
Foivos: yea can do it (there is a node query for memory slots) but it would be fairly hard to write into them (I think easier would be to extend ITS to allow it to write into it's memory on this branch, since we still use InterchainAddressTracker to be able to remove). Will skip for now unless we think we need to add the tests.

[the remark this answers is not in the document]
milapsheth: ok, yeah, fine to skip it

contracts/utils/ItsHubAddressTracker.sol
```diff
@@ -34,8 +38,10 @@ abstract contract ItsHubAddressTracker is IItsHubAddressTracker {
         bytes32 itsHubAddressMiddle_ = itsHubAddressMiddle;
         uint8 itsHubAddressSuffix_ = itsHubAddressSuffix;
         hubAddress = new string(65);
-        assembly {
+
+        assembly ('memory-safe') {
             mstore(add(hubAddress, 32), itsHubAddressPrefix_)
+            // The below writes 32 bytes into the middle slot and 1 bytes into the suffix slot, so it is done before the middle write.
```
milapsheth, line 44:
```suggestion
            // This writes the 1-byte suffix to the 65th position, but also 31 other bytes into the middle. Hence, the middle is written after this to prevent it from being overwritten.
```

test/InterchainTokenService.js
```diff
@@ -331,6 +331,19 @@ describe('Interchain Token Service', () => {
             ]);
         });
 
+        it('Should cleat previously set addresses', async () => {
```
milapsheth, line 334:
```suggestion
        it('Should clear previously set addresses', async () => {
```

test/InterchainTokenService.js
```diff
@@ -331,6 +331,19 @@ describe('Interchain Token Service', () => {
...
+            const operator = wallet.address;
+            const trustedChainName = 'ChainA';
+            const trustedAddress = wallet.address;
+
+            await expect(serviceTest.setTrustedAddress(trustedChainName, trustedAddress))
+                .to.emit(serviceTest, 'TrustedAddressSet')
+                .withArgs(trustedChainName, trustedAddress);
+
+            const params = defaultAbiCoder.encode(['address', 'string', 'string[]'], [operator, chainName, [trustedChainName]]);
+            await expect(serviceTest.setupTest(params)).to.emit(serviceTest, 'TrustedAddressRemoved').withArgs(trustedChainName);
```
milapsheth, line 344: :+1:

test/UtilsTest.js
```diff
@@ -256,6 +256,31 @@ describe('FlowLimit', async () => {
...
+        expect(await test.isTrustedChain(chainName)).to.equal(false);
+
+        await expect(test.setTrustedChain(chainName)).to.emit(test, 'TrustedChainSet').withArgs(chainName);
+
+        expect(await test.isTrustedChain(chainName)).to.equal(true);
+
+        await expect(test.removeTrustedChain(chainName)).to.emit(test, 'TrustedChainRemoved').withArgs(chainName);
+
+        expect(await test.isTrustedChain(chainName)).to.equal(false);
+    });
```
milapsheth, line 281: this should be tested on the ITS contract directly

milapsheth (approved):

test/UtilsTest.js
Foivos: it is tested in this section: https://github.com/axelarnetwork/interchain-token-service/blob/feat/only-hub-chains/test/InterchainTokenService.js#L697

PR #338, opened 2025-05-01
chore: remove set trusted address from test code

ahramy:
[AXE-9250](https://axelarnetwork.atlassian.net/browse/AXE-9250)

[AXE-9250]: https://axelarnetwork.atlassian.net/browse/AXE-9250?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/test/InterchainTokenService.js
+++ b/test/InterchainTokenService.js
@@ -333,19 +333,6 @@ describe('Interchain Token Service', () => {
             ]);
         });
 
-        it('Should clear previously set addresses', async () => {
-            const operator = wallet.address;
-            const trustedChainName = 'ChainA';
-            const trustedAddress = wallet.address;
-
-            await expect(serviceTest.setTrustedAddress(trustedChainName, trustedAddress))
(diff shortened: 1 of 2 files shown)


test/InterchainTokenService.js
```diff
@@ -333,19 +333,6 @@ describe('Interchain Token Service', () => {
             ]);
         });
 
-        it('Should clear previously set addresses', async () => {
```
milapsheth, line 336: This test is to test the migration in setup. It requires setting the old style addresses first

milapsheth (changes requested):

test/InterchainTokenService.js
ahramy: "This was specifically added to test the migration logic in setup hence the codecov failure after removing it, we should not remove it"

PR #340, opened 2025-05-01
refactor: use its message encoding for tests

ahramy:
[AXE-9171](https://axelarnetwork.atlassian.net/browse/AXE-9171)

[AXE-9171]: https://axelarnetwork.atlassian.net/browse/AXE-9171?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/test/AddressDerivation.js
+++ b/test/AddressDerivation.js
@@ -9,16 +9,10 @@ const {
 } = ethers;
 const { deployAll } = require('../scripts/deploy');
 const { approveContractCall } = require('../scripts/utils');
-const { getRandomBytes32, getSaltFromKey, isHardhat, getContractJSON, encodeITSPayload } = require('./utils');
+const { getRandomBytes32, getSaltFromKey, isHardhat, getContractJSON, encodeDeployInterchainToken } = require('./utils');
 const { create3DeployContract } = require('@axelar-network/axelar-gmp-sdk-solidity');
 const Token = getContractJSON('TestInterchainTokenStandard');
-const {
-    NATIVE_INTERCHAIN_TOKEN,
-    MESSAGE_TYPE_DEPLOY_INTERCHAIN_TOKEN,
-    MESSAGE_TYPE_RECEIVE_FROM_HUB,
-    ITS_HUB_ADDRESS,
-    ITS_HUB_CHAIN,
-} = require('./constants');
+const { NATIVE_INTERCHAIN_TOKEN, MESSAGE_TYPE_RECEIVE_FROM_HUB, ITS_HUB_ADDRESS, ITS_HUB_CHAIN } = require('./constants');
 
 if (isHardhat) {
     describe('Token Address Derivation [ @skip-on-coverage ]', () => {
@@ -63,14 +57,15 @@ if (isHardhat) {
                 const expectedTokenManagerAddress = '0x3B058fE7Ed045f56F0152AED1e8c5fbaE7e23C70';
 
                 const params = defaultAbiCoder.encode(['bytes', 'address'], [operator, expectedTokenAddress]);
-                const { payload } = encodeITSPayload(MESSAGE_TYPE_RECEIVE_FROM_HUB, sourceChain, [
-                    MESSAGE_TYPE_DEPLOY_INTERCHAIN_TOKEN,
+                const { payload } = encodeDeployInterchainToken(
+                    MESSAGE_TYPE_RECEIVE_FROM_HUB,
+                    sourceChain,
                     tokenId,
                     tokenName,
                     tokenSymbol,
                     tokenDecimals,
                     minter,
-                ]);
+                );
                 const commandId = await approveContractCall(gateway, ITS_HUB_CHAIN, ITS_HUB_ADDRESS, service.address, payload);
 
                 await expect(service.execute(commandId, ITS_HUB_CHAIN, ITS_HUB_ADDRESS, payload))
@@ -90,14 +85,15 @@ if (isHardhat) {
                 const expectedTokenManagerAddress = '0x89AF99D9373722De5F29ABbF706efeD020ae3E1F';
 
                 const params = defaultAbiCoder.encode(['bytes', 'address'], [operator, expectedTokenAddress]);
-                const { payload } = encodeITSPayload(MESSAGE_TYPE_RECEIVE_FROM_HUB, sourceChain, [
-                    MESSAGE_TYPE_DEPLOY_INTERCHAIN_TOKEN,
+                const { payload } = encodeDeployInterchainToken(
+                    MESSAGE_TYPE_RECEIVE_FROM_HUB,
+                    sourceChain,
                     tokenId,
                     tokenName,
                     tokenSymbol,
                     tokenDecimals,
                     minter,
-                ]);
+                );
                 const commandId = await approveContractCall(gateway, ITS_HUB_CHAIN, ITS_HUB_ADDRESS, service.address, payload);
 
                 await expect(service.execute(commandId, ITS_HUB_CHAIN, ITS_HUB_ADDRESS, payload))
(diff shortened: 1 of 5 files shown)


test/utils.js
```diff
@@ -187,6 +195,63 @@ function getContractJSON(contractName, artifactPath) {
...
+    };
+}
+
+/**
+ * Encodes an Interchain Token Service (ITS) message with wrapping.
+ * Supports single chain (returns object) or array of chains (returns array of objects).
+ *
+ * @param {number} wrapperType - Wrapper message type (e.g. RECEIVE_FROM_HUB, SEND_TO_HUB)
+ * @param {string|string[]} chain - Single chain name or array of chains
+ * @param {Array<any>} values - Values for the ITS payload (first item must be messageType)
```
milapsheth, line 241: methods should have a clean API and be safe to use. Using dynamic types and identifying errors at run time is not a great solution. Think about how you'd do this in Rust idiomatically to enforce safety

milapsheth: At the least, I'd prefer to see separate well-typed encoding methods like encodeInterchainTransfer, etc.

test/utils.js
```diff
@@ -187,6 +195,63 @@ function getContractJSON(contractName, artifactPath) {
...
+            return ['uint256', 'bytes32', 'bytes', 'bytes', 'uint256', 'bytes'];
+
+        case MESSAGE_TYPE_DEPLOY_TOKEN_MANAGER:
+            return ['uint256', 'bytes32', 'bytes', 'uint256'];
+
+        case MESSAGE_TYPE_REGISTER_TOKEN_METADATA:
+            return ['uint256', 'bytes', 'uint8'];
+
+        case INVALID_MESSAGE_TYPE:
+            return ['uint256', 'bytes32', 'bytes', 'uint256'];
```
milapsheth, line 220: this case doesn't make any sense to exist in a util

test/utils.js
```diff
@@ -187,6 +195,63 @@ function getContractJSON(contractName, artifactPath) {
...
+
+        default:
+            throw new Error(`Unknown ITS message type: ${messageType}`);
+    }
+}
+
+function wrapPayload(wrapperType, chain, payload) {
```
milapsheth, line 227: use a more meaningful name

milapsheth (changes requested):

test/utils.js
ahramy: Thanks for the comments, updated.

test/utils.js
ahramy: got it, updated

test/utils.js
ahramy: updated

test/AddressDerivation.js
```diff
@@ -90,17 +85,18 @@ if (isHardhat) {
...
                 const params = defaultAbiCoder.encode(['bytes', 'address'], [operator, expectedTokenAddress]);
-                const payload = defaultAbiCoder.encode(
-                    ['uint256', 'bytes32', 'string', 'string', 'uint8', 'bytes'],
-                    [MESSAGE_TYPE_DEPLOY_INTERCHAIN_TOKEN, tokenId, tokenName, tokenSymbol, tokenDecimals, minter],
-                );
-                const wrappedPayload = defaultAbiCoder.encode(
-                    ['uint256', 'string', 'bytes'],
-                    [MESSAGE_TYPE_RECEIVE_FROM_HUB, sourceChain, payload],
+                const { payload } = encodeDeployInterchainToken(
+                    MESSAGE_TYPE_RECEIVE_FROM_HUB,
```
milapsheth, line 89: This argument is out of place for what the function is trying to do. Look at how we have structured the ItsMessage vs ItsHubMessage in stellar ITS. Similar I recommend having two layers, encode* function for the underlying ITS message and then another encode method for send to hub or receive from hub

test/utils.js
```diff
@@ -187,6 +194,65 @@ function getContractJSON(contractName, artifactPath) {
     }
 }
 
+function encodeDeployInterchainToken(wrapperType, chain, tokenId, name, symbol, decimals, minter, operator = null) {
+    const values = operator
```
milapsheth, line 198: This doesn't make sense. There's no separate operator field on this msg, only a minter?

test/utils.js
```diff
@@ -187,6 +194,65 @@ function getContractJSON(contractName, artifactPath) {
...
+
+function encodeAndHashPayload(wrapperType, chain, payload) {
+    const wrappedPayload = defaultAbiCoder.encode(['uint256', 'string', 'bytes'], [wrapperType, chain, payload]);
+    return {
+        payload: wrappedPayload,
+        payloadHash: keccak256(wrappedPayload),
+    };
+}
+
+function encodeITSWrappedPayload(wrapperType, chain, payload) {
```
milapsheth, line 248: Bad naming, should be meaningful like encodeHubMessage

test/utils.js
```diff
@@ -187,6 +194,65 @@ function getContractJSON(contractName, artifactPath) {
...
+
+function encodeRegisterTokenMetadata(tokenAddress, decimals) {
+    const payload = defaultAbiCoder.encode(['uint256', 'bytes', 'uint8'], [MESSAGE_TYPE_REGISTER_TOKEN_METADATA, tokenAddress, decimals]);
+    return {
+        payload,
+        payloadHash: keccak256(payload),
+    };
+}
+
+function encodeAndHashPayload(wrapperType, chain, payload) {
```
milapsheth, line 240: Bad naming, encode how/what?

test/utils.js
```diff
@@ -187,6 +194,65 @@ function getContractJSON(contractName, artifactPath) {
...
+    const wrappedPayload = defaultAbiCoder.encode(['uint256', 'string', 'bytes'], [wrapperType, chain, payload]);
+    return {
+        payload: wrappedPayload,
+        payloadHash: keccak256(wrappedPayload),
+    };
+}
+
+function encodeITSWrappedPayload(wrapperType, chain, payload) {
+    if (Array.isArray(chain)) {
```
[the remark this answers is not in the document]
ahramy: got it. make sense, updated.

test/utils.js
ahramy: This was from the testcase: https://github.com/axelarnetwork/interchain-token-service/blob/35c82fff1fce6fbc0b8fbbd155a02a13e2f728b2/test/InterchainTokenService.js#L252
and I was trying to preserve the original case. but since there’s no separate operator field, I removed operator.

test/utils.js
```diff
@@ -195,61 +194,45 @@ function getContractJSON(contractName, artifactPath) {
...
-        default:
-            throw new Error(`Unknown ITS message type: ${messageType}`);
-    }
+function encodeDeployTokenManagerMessage(tokenId, address, salt) {
+    return defaultAbiCoder.encode(['uint256', 'bytes32', 'bytes', 'uint256'], [MESSAGE_TYPE_DEPLOY_TOKEN_MANAGER, tokenId, address, salt]);
 }
 
-function wrapPayload(wrapperType, chain, payload) {
-    const wrappedPayload = defaultAbiCoder.encode(['uint256', 'string', 'bytes'], [wrapperType, chain, payload]);
+function encodeHubMessage(wrapperType, chain, message) {
```
milapsheth, line 215: `wrapperType` bad naming. This is still not as safe since someone could pass any value as the type. Instead create dedicated methods `encodeSendHubMessage` and similarly for destination

test/utils.js
ahramy: Yeah, I meant that there was actually an additional operator parameter - the test case had minter and operator as separate parameters. But removed the operator since it would be the same

test/utils.js
ahramy: got it. updated

test/InterchainTokenFactory.js
```diff
@@ -139,14 +144,8 @@ describe('InterchainTokenFactory', () => {
...
-            const { payload, payloadHash } = encodeITSPayload(MESSAGE_TYPE_SEND_TO_HUB, destinationChain, [
-                MESSAGE_TYPE_DEPLOY_INTERCHAIN_TOKEN,
-                tokenId,
-                name,
-                symbol,
-                decimals,
-                '0x',
-            ]);
+            const message = encodeDeployInterchainTokenMessage(tokenId, name, symbol, decimals, '0x');
+            const { payload, payloadHash } = encodeHubMessage(MESSAGE_TYPE_SEND_TO_HUB, destinationChain, message);
```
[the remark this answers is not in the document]
ahramy: updated!

PR #341, opened 2025-05-05
refactor(its): remove metadata version

ahramy:
[AXE-9108](https://axelarnetwork.atlassian.net/browse/AXE-9108)

[AXE-9108]: https://axelarnetwork.atlassian.net/browse/AXE-9108?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/contracts/InterchainTokenService.sol
+++ b/contracts/InterchainTokenService.sol
@@ -100,7 +100,7 @@ contract InterchainTokenService is
     /**
      * @dev Latest version of metadata that's supported.
      */
-    uint32 internal constant LATEST_METADATA_VERSION = 0;
+    uint32 internal constant METADATA_CONTRACT_CALL = 0;
 
     /**
      * @dev Chain name where ITS Hub exists. This is used for routing ITS calls via ITS hub.
@@ -576,7 +576,7 @@ contract InterchainTokenService is
         bytes calldata metadata,
         uint256 gasValue
     ) external payable whenNotPaused {
-        (, bytes memory data) = _decodeMetadata(metadata);
+        bytes memory data = _decodeMetadata(metadata);
 
         _interchainTransfer(tokenId, destinationChain, destinationAddress, amount, data, gasValue);
     }
@@ -605,7 +605,7 @@ contract InterchainTokenService is
     ) external payable whenNotPaused {
         amount = _takeToken(tokenId, sourceAddress, amount, true);
 
-        (, bytes memory data) = _decodeMetadata(metadata);
+        bytes memory data = _decodeMetadata(metadata);
 
         _transmitInterchainTransfer(tokenId, sourceAddress, destinationChain, destinationAddress, amount, data, msg.value);
     }
@@ -1029,18 +1029,17 @@ contract InterchainTokenService is
      * @notice Decodes the metadata into a version number and data bytes.
      * @dev The function expects the metadata to have the version in the first 4 bytes, followed by the actual data.
      * @param metadata The bytes containing the metadata to decode.
-     * @return version The version number extracted from the metadata.
      * @return data The data bytes extracted from the metadata.
      */
-    function _decodeMetadata(bytes calldata metadata) internal pure returns (uint32 version, bytes memory data) {
-        if (metadata.length < 4) return (0, data);
+    function _decodeMetadata(bytes calldata metadata) internal pure returns (bytes memory data) {
+        if (metadata.length < 4) return data;
 
         uint32 versionUint = uint32(bytes4(metadata[:4]));
-        if (versionUint > LATEST_METADATA_VERSION) revert InvalidMetadataVersion(versionUint);
+        // This check is maintained for backward compatibility with existing integrations
+        // that expect a specific metadata format with METADATA_CONTRACT_CALL version
+        if (versionUint != METADATA_CONTRACT_CALL) revert InvalidMetadataVersion(versionUint);
 
-        version = versionUint;
-        if (metadata.length == 4) return (version, data);
-
+        if (metadata.length == 4) return data;
         data = metadata[4:];
     }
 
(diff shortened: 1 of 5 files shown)


contracts/InterchainTokenService.sol
```diff
@@ -592,9 +576,9 @@ contract InterchainTokenService is
         bytes calldata metadata,
         uint256 gasValue
     ) external payable whenNotPaused {
-        (IGatewayCaller.MetadataVersion metadataVersion, bytes memory data) = _decodeMetadata(metadata);
+        (, bytes memory data) = _decodeMetadata(metadata);
```
ahramy, line 579: I did not remove `_decodeMetadata` to maintain compatibility in case there are any external function calls involving metadata (e.g. Squid?) @milapsheth I'm not sure whether we could completely remove `_decodeMetata` or not. Could you let me know where I should look to confirm this?

milapsheth: We need to support the existing format since the only way to make transfers with contract call requires using metadata. However, _decodeMetadata should instead enforce that only the contract call metadata type is being used and it doesn't need to return the metadata value anymore

contracts/InterchainTokenService.sol
```diff
@@ -1054,14 +1032,13 @@ contract InterchainTokenService is
      * @return version The version number extracted from the metadata.
      * @return data The data bytes extracted from the metadata.
      */
-    function _decodeMetadata(bytes calldata metadata) internal pure returns (IGatewayCaller.MetadataVersion version, bytes memory data) {
-        if (metadata.length < 4) return (IGatewayCaller.MetadataVersion.CONTRACT_CALL, data);
+    function _decodeMetadata(bytes calldata metadata) internal pure returns (uint32 version, bytes memory data) {
```
milapsheth, line 1035: No need to return the metadata version anymore

contracts/InterchainTokenService.sol
```diff
@@ -1054,14 +1032,13 @@ contract InterchainTokenService is
...
+        if (metadata.length < 4) return (0, data);
 
         uint32 versionUint = uint32(bytes4(metadata[:4]));
```
milapsheth, line 1038: Add a comment on why this check still exists for backcompat

contracts/utils/GatewayCaller.sol
```diff
@@ -29,30 +29,24 @@ contract GatewayCaller is IGatewayCaller {
      * @param destinationChain The target chain where the contract will be called
      * @param destinationAddress The address of the contract to be called on the destination chain
      * @param payload The data payload for the transaction
-     * @param metadataVersion The version of metadata to be used
      * @param gasValue The amount of gas to be paid for the cross-chain message. If this is 0, then gas payment is skipped. `msg.value` must be at least gasValue.
      */
     function callContract(
```
milapsheth, line 34: In a follow up PR, ITS can inherit the gateway caller instead of calling it as an external contract, if ITS bytecode size is within limit. Alternatively the contract can just be removed altogether

contracts/utils/GatewayCaller.sol
```diff
@@ -29,30 +29,24 @@ contract GatewayCaller is IGatewayCaller {
...
-                    destinationAddress,
-                    payload,
-                    // solhint-disable-next-line avoid-tx-origin
-                    tx.origin
-                );
-            } else {
-                revert InvalidMetadataVersion(uint32(metadataVersion));
-            }
+            // slither-disable-next-line arbitrary-send-eth
+            gasService.payNativeGasForContractCall{ value: gasValue }(
```
milapsheth, line 42: In a separate PR, switch to using payGas instead which is just the newer more forward compatible method

milapsheth (approved):

contracts/InterchainTokenService.sol
```diff
@@ -291,7 +291,7 @@ contract InterchainTokenService is
 
         emit TokenMetadataRegistered(tokenAddress, decimals);
```
milapsheth, line 292: Add change set file

contracts/InterchainTokenService.sol
ahramy: make sense, updated

contracts/InterchainTokenService.sol
ahramy: updated

contracts/InterchainTokenService.sol
```diff
@@ -1054,14 +1032,13 @@ contract InterchainTokenService is
      * @return version The version number extracted from the metadata.
      * @return data The data bytes extracted from the metadata.
      */
-    function _decodeMetadata(bytes calldata metadata) internal pure returns (IGatewayCaller.MetadataVersion version, bytes memory data) {
-        if (metadata.length < 4) return (IGatewayCaller.MetadataVersion.CONTRACT_CALL, data);
+    function _decodeMetadata(bytes calldata metadata) internal pure returns (uint32 version, bytes memory data) {
+        if (metadata.length < 4) return (0, data);
 
         uint32 versionUint = uint32(bytes4(metadata[:4]));
         if (versionUint > LATEST_METADATA_VERSION) revert InvalidMetadataVersion(versionUint);
```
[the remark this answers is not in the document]
ahramy: make sense, updated

contracts/utils/GatewayCaller.sol
ahramy: got it, created a ticket: https://axelarnetwork.atlassian.net/browse/AXE-9291

contracts/utils/GatewayCaller.sol
ahramy: got it, created a ticket: https://axelarnetwork.atlassian.net/browse/AXE-9292

contracts/InterchainTokenService.sol
ahramy: added changeset

PR #342, opened 2025-05-06
chore(its): use payGas instead of payNativeGasForContractCall

ahramy:
[AXE-9291](https://axelarnetwork.atlassian.net/browse/AXE-9291)

### DIFF
--- a/contracts/utils/GatewayCaller.sol
+++ b/contracts/utils/GatewayCaller.sol
@@ -39,13 +39,16 @@ contract GatewayCaller is IGatewayCaller {
     ) external payable override {
         if (gasValue > 0) {
             // slither-disable-next-line arbitrary-send-eth
-            gasService.payNativeGasForContractCall{ value: gasValue }(
+            gasService.payGas{ value: gasValue }(
                 address(this),
                 destinationChain,
                 destinationAddress,
                 payload,
+                0,
+                false,
                 // solhint-disable-next-line avoid-tx-origin
-                tx.origin
+                tx.origin,
+                ''
             );
         }
 


contracts/utils/GatewayCaller.sol
```diff
@@ -38,14 +38,21 @@ contract GatewayCaller is IGatewayCaller {
...
+            gasService.payGas{ value: gasValue }(
                 address(this),
                 destinationChain,
                 destinationAddress,
                 payload,
+                executionGasLimit,
+                estimateOnChain,
                 // solhint-disable-next-line avoid-tx-origin
-                tx.origin
+                tx.origin,
```
milapsheth, line 54: create a refund address variable?

contracts/utils/GatewayCaller.sol
```diff
@@ -38,14 +38,21 @@ contract GatewayCaller is IGatewayCaller {
         uint256 gasValue
     ) external payable override {
         if (gasValue > 0) {
+            uint256 executionGasLimit = 0;
+            bool estimateOnChain = false;
```
milapsheth, line 42:
```suggestion
            // Gas for the ITS msg must be estimated off-chain
            bool estimateOnChain = false;
            // No need to set the gas limit since it's not being estimated on-chain
            uint256 executionGasLimit = 0;
```
explain the reasoning for the configuration

milapsheth (approved):

contracts/utils/GatewayCaller.sol
ahramy: updated

PR #352, opened 2025-07-29
fix(interchain-token-factory)!: remove deprecated ITS factory methods

milapsheth:
[AXE-9851]

[AXE-9851]: https://axelarnetwork.atlassian.net/browse/AXE-9851?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


test/InterchainTokenFactory.js
```diff
@@ -7,7 +7,7 @@ const {
     getContractAt,
     Wallet,
     constants: { AddressZero, HashZero },
-    utils: { defaultAbiCoder, keccak256, toUtf8Bytes, arrayify },
+    utils: { defaultAbiCoder, keccak256, toUtf8Bytes },
```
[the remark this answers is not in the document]
milapsheth: invalid comment

test/InterchainTokenFactory.js
```diff
@@ -314,267 +263,233 @@ describe('InterchainTokenFactory', () => {
             await checkRoles(tokenManager, minter);
         });
 
-        it('Should initiate a remote interchain token deployment with the same minter', async () => {
+        it('Should revert when initiating a remote interchain token deployment with minter as zero address and no destination minter', async () => {
```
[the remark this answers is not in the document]
milapsheth: i'm just emphasizing that no destination minter is set. There is no additional error being tested there. Just want to test minter as address(0) when both destination minter is set or not set

contracts/InterchainTokenFactory.sol
```diff
@@ -251,17 +251,18 @@ contract InterchainTokenFactory is IInterchainTokenFactory, Multicall, Upgradabl
...
-     * no additional minter is set on the token. Reverts if the minter does not have mint permission for the token.
+     * @param minter The minter of the interchain token on the current chain. If the token has no minter other than ITS, then no minter can be set on the interchain token on the destination chain, and `deployRemoteInterchainToken` must be used instead.
      * @param destinationChain The name of the destination chain.
-     * @param destinationMinter The minter address to set on the deployed token on the destination chain. This can be arbitrary bytes
-     * since the encoding of the account is dependent on the destination chain. If this is empty, then the `minter` of the token on the current chain
+     * @param destinationMinter The address to receive the minter and operator role of the token on the destination chain. This can be arbitrary bytes
+     * since the encoding of the account is dependent on the destination chain. The `minter` must have approved a custom `destinationMinter` via `approveDeployRemoteInterchainToken` beforehand. If this is empty, then the `minter` of the token on the current chain
```
ahramy, line 265: If we're using ITS as a minter here, we can mention that the ITS address doesn't need to be approved?

ahramy: Yup, the comment is for `deployRemoteInterchainTokenWithMinter`. Just to confirm, are we going to allow a user to approve and use ITS as a minter for the `deployRemoteInterchainTokenWithMinter` function?

milapsheth: the user can choose arbitrary bytes as the `destinationMinter`. There's no point of checking that. That address is on the destination chain, so we can't even check it (it might be a valid account on that chain and not be ITS even if that address on the current chain corresponds to ITS)

ahramy (approved):

